PR #37, opened 2019-12-16
Make sure mutexes are consistently used in polytracker

### LINKED ISSUE #25: Mutexes aren't used consistently in the polytracker runtime implementation in `dfsan.cc`
In particular, look for the uses of `function_to_bytes` and `union_table` within `polytracker/dfsan/dfsan_rt/dfsan/dfsan.cc`. Each of these variables has an accompanying `std::mutex`, but the mutex is not taken everywhere the variables are touched.

Additionally, it would be more robust to use `std::lock_guard<std::mutex>` or `std::unique_lock<std::mutex>` instead of manually calling `.lock()` and `.unlock()` on the mutex variables.

bradlarsen:
In the future, we might consider using just a single mutex for all polytracker internals, since once you start dealing with multiple locks, you have to worry about deadlock.

Fixes #25.


ThatsNotVeryCashMoneyOfYou:
It looks good 👍 but has conflicts again lol sorry. If you fix it again, you can automerge it.

bradlarsen:
Let me fix the conflicts...

carsonharmon:
This has bit rotted

Issue #18, opened 2019-12-09
Runtime CFG Tracing


ESultanik:
Track the "called from" function for each function emitted in the PolyTracker output.
```json
{
  "dfs$ensure_solid_xref": {
    "called_from": [list of function names that called dfs$ensure_solid_xref],
    "input_bytes": {
      "POLYTRACK filename": [
        2276587,
        2276588
      ]
    }
  },
```
Note that this output format assumes the modifications from #17 have been implemented.
Alternatively, if it is easier to implement, we could list the set of functions each function _calls_:
```json
{
  "dfs$ensure_solid_xref": {
    "called": [list of function names that were called by dfs$ensure_solid_xref],
    "input_bytes": {
      "POLYTRACK filename": [
        2276587,
        2276588
      ]
    }
  },
```

Issue #17, opened 2019-12-09
Add source information to the JSON output


ESultanik:
The output should include the name of the input file from which the bytes originated. This will also future-proof the format for if/when PolyTracker enables tracking from multiple files (or sockets?) at once.
```json
{
  "dfs$ensure_solid_xref": {
    "input_bytes": {
      "POLYTRACK filename": [
        2276587,
        2276588
      ]
    }
  },
⋮
```

ThatsNotVeryCashMoneyOfYou:
I'm going to implement this via repurposing our bitfield to keep track of what taint source its from. This will allow us to track 8 taint sources. If we see more than one bit flipped, we know its influenced from more than one taint source.

ThatsNotVeryCashMoneyOfYou:
This might be a bit tricker to support multiple taint sources, i.e. right now our addressing schema is based around taint labels corresponding to offsets, this doesnt work well if we have multiple files.

notBD:
This has been done since dec/jan

PR #38, opened 2019-12-17
Carson/build runtime cfg

ThatsNotVeryCashMoneyOfYou:
This PR addresses issues #17  and #18 and has a large refactor of how we manage taint information. 

1. We are able to track where taint comes from for sources, and now it wont take much effort to track multiple sources at once 
2. The refactor helps with locking issues 
3. Json scheme is changed, see where we modify output_json in dfsan.cc 
4. Will refactor dfsan itself to deal with locking etc. There is no need for us to have this many global structures with locks like this, its messy and could lead to issues. 
5. There is a runtime CFG that is made by examining the function that called the current one in the dfsan_func_entry.


ThatsNotVeryCashMoneyOfYou:
@ESultanik

PR #75, opened 2020-06-17
Carson/basic blocks

carsonharmon:
The PR completely changes our build pipeline and might have some other additions i forgot about. 

Some highlights: 

We now vendor our own libcxx and libcxx bitcode, this allows us to force targets to build our libcxx and instrument the full libcxx bitcode without headache. 

We replaced polyclang which was a wrapper around clang with polybuild, a python wrapper around gclang/gllvm 

The pipeline works as follows - Compile everything with gllvm, if cxx, make sure they use our cxx lib. Store all the build artifacts and build commands into a manifest. Find the target binary, extract the bitcode. Then you can instrument the bitcode using the commands in the readme. When we instrument we mostly don't touch the ABI, because that breaks things, but it turns out just going through the standard lib seems to work on the testcases.  

${CXX} --instrument-bitcode -f clean_bitcode.bc -o track_target --libs <libs here> 

This PR also introduces a helper script which will help show differences between track lists. 

We also have C and CXX test cases for polytracker running in pytest, the test cases are minimal but show that we are actually propagating taint through the cxx standard library etc. 

Slight modification to the instrumentation pass to prioritize custom definitions over discarded definitions 

This is like 2ish weeks of work, so i dont know what else is in here. There is some tech debt in polybuild, but the dockerfiles seem to build okay. Please review lol.


polybuild/polybuild.py
```diff
@@ -0,0 +1,358 @@
...
+        compile_command += ["-Wl,--whole-archive", rt_dir, "-Wl,--no-whole-archive", source_dir]
+        compile_command += ["-ldl", "-lrt"]
+        compile_command.append("-lstdc++")
+        for lib in libs:
+            if ".a" not in lib and ".o" not in lib:
+                compile_command.append("-l" + lib)
+            else:
+                compile_command.append(lib)
+        command = shlex.split(" ".join(shlex.quote(arg) for arg in compile_command))
+        ret_code = subprocess.call(command)
```
ESultanik, line 110: I don't think the call to `shlex.split` is necessary. Instead, I think you should be able to just do `subprocess.call(compile_command)`.

polybuild/polybuild.py
```diff
@@ -0,0 +1,358 @@
...
+        for arg in argv[1:]:
+            compile_command.append(arg)
+        is_linking = self.poly_is_linking(argv)
+        if is_linking:
+            # If its cxx, link in our c++ libs
+            if self.meta.is_cxx:
+                compile_command += ["-lc++", "-lc++abipoly", "-lc++abi", "-lpthread"]
+
+        command = shlex.split(" ".join(shlex.quote(arg) for arg in compile_command))
+        res = subprocess.call(command)
```
ESultanik, line 174: Same here, I don't think you need `shlex.split` if you pass `compile_command` directly to `subprocess.call`. _However!_ If you do that, then you might need to split up some of the elements of the compile command. For example, I think `compile_command.append("-I" + self.meta.compiler_dir + "/../cxx_libs/include/c++/v1/")` would have to become `compile_command += ["-I", self.meta.compiler_dir, "/../cxx_libs/include/c++/v1/"]`.

polybuild/polybuild.py
carsonharmon: I will give it a try, but originally when i was doing this doing the subprocess call had issues when passing a string with additional double quotes inside of it. for example -DSOME_VAR="whatever", it would remove them. So I iterated through and created the new array by doing a join/split, but maybe i can try again.

polybuild/polybuild.py
```diff
@@ -0,0 +1,358 @@
...
+    get-bc -b <target_binary> 
+
+    OPTIONAL: Link multiple bitcode files together with llvm-link
+
+    Instrument that whole program bitcode file by invoking 
+    polybuild --instrument-bitcode -f <bitcode_file.bc> -o <output_file> 
+    """
+    )
+    parser.add_argument("--instrument-bitcode", action="store_true", help="Specify to add polytracker instrumentation")
+    parser.add_argument("--input-file", "-f", type=str, default=None, help="Path to the whole program bitcode file")
```
ESultanik, line 235: If you take out `default=None,` then `argparse` will automatically fail with a nice error message if it is omitted and you can remove your manual check below.

polybuild/polybuild.py
carsonharmon: I don't think it splits it up that way, when you do the "-I" + string + .... and include it into the list it gets treated as one arg I believe. At least thats what I saw from my command line logging output. I can check again

polybuild/polybuild.py
carsonharmon: can you explain this further? So when I actually decide to parse the args, if there is no default value, it just fails and says i need it? that sounds nice, ill do that.

polybuild/polybuild.py
carsonharmon: Done!

polybuild/polybuild.py
carsonharmon: Done!

polybuild/polybuild.py
```diff
@@ -0,0 +1,345 @@
...
+
+
+"""
+Store a build artifact to the artifact storage via copy
+"""
+
+
+def store_artifact(file_path, artifact_path) -> bool:
+    filename = get_file_name(file_path)
+    artifact_file_path = artifact_path + "/" + filename
```
ESultanik, line 182:
```suggestion
    artifact_file_path = os.path.join(artifact_path, filename)
```

polybuild/polybuild.py
```diff
@@ -0,0 +1,345 @@
...
+    # Check if its an absolute path
+    if file_path[0] != "/":
+        cwd = os.getcwd()
+        targ_path = cwd + "/" + file_path
```
ESultanik, line 186:
```suggestion
        targ_path = os.path.join(cwd, file_path)
```

polybuild/polybuild.py
```diff
@@ -0,0 +1,345 @@
...
+        return False
+    ret = subprocess.call(["cp", targ_path, artifact_file_path])
+    if ret != 0:
+        print(f"Error! failed to store {targ_path}, error was {ret}")
+        print(f"Artifact path: {artifact_file_path}")
+        return False
+    return True
+
+
+def get_file_name(file_path: str) -> str:
```
ESultanik, line 200: Not exactly sure what this function is meant for, but you might be able to replace it with [`os.path.split`](https://docs.python.org/3/library/os.path.html#os.path.split) or [`os.path.basename`](https://docs.python.org/3/library/os.path.html#os.path.basename)

carsonharmon: I think you are correct, ill investigate

polybuild/polybuild.py
carsonharmon: Removed shlex so that should all be good.

tests/test_polytracker.py
```diff
@@ -0,0 +1,120 @@
+import pytest
+import os
+from polyprocess import PolyProcess
+
+cwd: str = os.getcwd()
```
ESultanik, line 5: You could replace these paths with something like this:
```python
TEST_DIR = os.path.realpath(os.path.dirname(__file__))
BIN_DIR = os.path.join(TEST_DIR, 'bin')
TEST_RESULTS_DIR = os.path.join(BIN_DIR, 'test_results')
BITCODE_DIR = os.path.join(test_dir, 'bitcode')
```

carsonharmon: That sounds like a good idea, ill do that

ESultanik: The first line will give you the absolute path to the directory in which the current script is located, regardless of the current working directory. This lets you call the code from a different CWD. Using `os.path.join` also makes the code portable, because it will handle Windows paths.

tests/test_polytracker.py
```diff
@@ -0,0 +1,120 @@
...
+
+This runs before any test is executed
+"""
+
+
+@pytest.fixture(scope="session", autouse=True)
+def setup_targets():
+    # Check if bin dir exists
+    if os.path.exists(bin_dir):
+        os.system("rm -r " + bin_dir)
```
ESultanik, line 22: This will fail if the path to `bin_dir` contains a space. You should probably change all of the `os.system` calls in here to `subprocess.call([arg1, arg2, arg3, ...])`.

carsonharmon: Ahh yeah, I forgot I have os.system still in the test files. I will replace it all with subprocess.call like I did in polybuild

ESultanik, line 22: You will need to add `import subprocess` above.
```suggestion
        subprocess.call(["rm", "-r", bin_dir])
```

carsonharmon: Done!

PR #81, opened 2020-07-22
PolyTracker Python and Installation Refactor

ESultanik:
1. Creates a `polytracker` Python package
2. Migrated PolyTracker JSON output parsing code from [PolyMerge](https://github.com/trailofbits/polyfile/blob/master/polymerge/polytracker.py). This will allow any Python script that uses `import polytracker` to parse PolyTracker JSON output into convenient Python objects without having to know the JSON schema. This will also allow for seamlessly changing the output format to something other than JSON (_e.g._, sqlite).
3. Moved `polyclang/polytracker.h` to `polytracker/polytracker.h` since `polyclang` doesn't really exist anymore.
4. Updated the version number and version numbering scheme in `polytracker/polytracker.h`
5. `setup.py` now automatically parses `polytracker/polytracker.h` and sets the Python package's version number in parity with the C++ version number.
6. Cleaned up `polybuild.py` to use `os.path.join` instead of manual path concatenation.
7. Changed the `CMake` installation directory structure
   * from:
      * `${CMAKE_INSTALL_PREFIX}/bin/polytracker/bin/polybuild.py`
      * `${CMAKE_INSTALL_PREFIX}/bin/polytracker/bin/polybuild++.py`
      * and `${CMAKE_INSTALL_PREFIX}/bin/polytracker` for everything else
   * to:
      * `${CMAKE_INSTALL_PREFIX}/bin/polybuild`
      * `${CMAKE_INSTALL_PREFIX}/bin/polybuild++`
      * and `${CMAKE_INSTALL_PREFIX}/share/polytracker` for everything else
      * with the `polybuild` scripts in the `bin` directory being symlinks to their actual install location under `/share/polytracker/`.
8. Added a GitHub action to automatically upload the Python package to PyPI on each release.
9. Cleaned up the Python 3.7 install in the `Dockerfile`
10. Added `polybuild` and `polybuild++` to the `$PATH` in the `Dockerfile`


polytracker/cfg.py
```diff
@@ -0,0 +1,177 @@
...
+    TypeVar,
+)
+
+import graphviz
+import networkx as nx
+
+N = TypeVar("N")
+
+
+class DiGraph(nx.DiGraph, Generic[N]):
```
carsonharmon, line 25: Could you explain this python to me? So you defined type N which I guess is the "Node" type, and you are using it to be a generic DiGraph right

polytracker/cfg.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def descendants(self, node) -> FrozenSet[N]:
+        return frozenset(nx.dfs_successors(self, node).keys())
+
+    @property
+    def dominator_forest(self) -> "DAG[N]":
+        if self._dominator_forest is not None:
+            return self._dominator_forest
+        self._dominator_forest = DAG()
+        for root in self.roots:
```
carsonharmon, line 66: Does this property access `self.roots` trigger the function above? (def roots)

polytracker/polytracker.py
```diff
@@ -0,0 +1,149 @@
...
+    for function_name, function_data in polytracker_json_obj.items():
+        if isinstance(function_data, dict) and "called_from" in function_data:
+            # this is the second version of the output format
+            return parse_format_v2(polytracker_json_obj)
+        else:
+            return parse_format_v1(polytracker_json_obj)
+    return parse_format_v1(polytracker_json_obj)
+
+
+@polytracker_version(0, 0, 1, "")
```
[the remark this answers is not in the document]
ESultanik: It registers the following function in a global dict that is used for figuring out which function is responsible for parsing which version of PolyTracker JSON output.

.github/workflows/dockerimage.yml
```diff
@@ -37,9 +37,9 @@ jobs:
           --exclude 'polytracker/src/dfsan_rt/interception/*'
       - name: Python lint/typecheck 
         run: |
-          black --check polybuild polyprocess tests --line-length=127
-          mypy --ignore-missing-imports polybuild polyprocess tests
-      - name: Build the base image 
+          black --check polytracker tests --exclude '/(polytracker/src|polytracker/scripts)/' --line-length=127
```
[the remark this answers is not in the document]
ESultanik: It's a bit of a hack. So, in Python, since we are now defining a `polytracker` package, it's standard practice to put that package in the `polytracker` subdirectory. But we already had our C++ code in there, which _also_ includes some _other_ Python scripts that are not included in the `polytracker` package (which we prevent by not including `__init__.py` files in their directories). Some of those scripts actually fail the `black` linting, which is why I just blanket exclude those directories. Note that `polybuild.py` is now included in the `scripts` directory, which is _excluded_ from this `black` check. So just keep in mind that `polybuild.py` will no longer be linted in CI. We should probably update all of the .py files in the `scripts` dir to satisfy `black` so we can remove that exclusion.

polytracker/cfg.py
ESultanik: This is solely for type hinting. It's basically the equivalent of this C++:
```c++
typedef <typename N>
class DiGraph : public nx::DiGraph { ... }
```
You can then use `N` within type hints through the rest of the class, like this:
https://github.com/trailofbits/polytracker/blob/e0af394394ea7c22c3e7b637230c3c7436c0529c/polytracker/cfg.py#L55-L56
And when you create an instance of that class, you can optionally provide a type for `N`, like I do here:
https://github.com/trailofbits/polytracker/blob/e0af394394ea7c22c3e7b637230c3c7436c0529c/polytracker/cfg.py#L157
So a `DiGraph` can have nodes of any type. But a `CFG`, which is a subclass of `DiGraph` specifically has nodes of type `FunctionInfo`. However, you could declare a DiGraph with _any_ type:
```python
graph: DiGraph[int] = DiGraph()
graph.add_node(1) # This is fine
graph.add_node('foo') # This will work but trigger a mypy/type hinting error
for root in graph.roots:
    bar: float = root # This will also cause a mypy/type hinting error
```

polytracker/cfg.py
ESultanik: Correct!
```python
class Foo:
    _bar = 1337

    @property
    def bar(self):
        return self._bar

    @bar.setter
    def bar(self, value):
        self._bar = value

>>> foo = Foo()
>>> print(foo.bar)
1337
>>> foo.bar = 0
>>> print(foo.bar)
0
```
If you don't define a `@bar.setter`, then you will get a runtime error if you try and set it. This is useful if you want to have member variables that are effectively read only.

PR #91, opened 2020-08-15
Split basic blocks

ESultanik:
This adds a pass immediately before the dfsan instrumentation that ensures each basic block contains at most one function call or one conditional branch. This means that each function call is always immediately followed by an unconditional branch to another basic block.

### DIFF
--- a/polytracker/src/dfsan_pass/BBSplittingPass.cpp
+++ b/polytracker/src/dfsan_pass/BBSplittingPass.cpp
@@ -5,21 +5,71 @@
  *      Author: Evan Sultanik, Trail of Bits
  */
 
+#include <iostream>
+
+#include "llvm/IR/InstrTypes.h"
+#include "llvm/IR/Instructions.h"
+
 #include "polytracker/bb_splitting_pass.h"
 
 using llvm::BasicBlock;
+using llvm::BranchInst;
+using llvm::CallInst;
+using llvm::dyn_cast;
+using llvm::Function;
+using llvm::Instruction;
+using llvm::isa;
+using llvm::TerminatorInst;
 
 namespace polytracker {
 
-bool BBSplittingPass::runOnBasicBlock(BasicBlock &F) {
+bool BBSplittingPass::analyzeBasicBlock(BasicBlock &basicBlock) const {
+  bool modified = false;
+  BasicBlock *bb = &basicBlock;
+
+  for (Instruction *inst = &bb->front(); inst && !isa<TerminatorInst>(inst);
+       inst = inst->getNextNode()) {
+    if (auto call = dyn_cast<CallInst>(inst)) {
+      // Is the call immediately followed by an unconditional branch?
+      // if so, that's the only case that is okay:
+      Instruction *next = inst->getNextNode();
+      if (auto branch = dyn_cast<BranchInst>(next)) {
+        if (branch->isUnconditional()) {
+          // The next instruction after the call is an unconditional branch,
+          // so it's okay to leave it.
+          continue;
+        }
+      }
+      // We need to split this BB into a new one after the call
+      modified = true;
+      bb = bb->splitBasicBlock(next);
+      std::cout << "Splitting basic block ";
+      if (bb->hasName()) {
+        std::cout << bb->getName().data();
+      }
+      std::cout << " after call to "
+                << call->getCalledFunction()->getName().data() << std::endl;
+    }
+    return modified;
+  }
+
   return false;
+}
+
+bool BBSplittingPass::runOnFunction(Function &function) {
+  bool ret = false;
+  for (auto &bb : function.getBasicBlockList()) {
+    ret = analyzeBasicBlock(bb) || ret;
+  }
+  return ret;
 }
 
 char BBSplittingPass::ID = 0;
 
-};
+}; // namespace polytracker
 
-
-static llvm::RegisterPass<polytracker::BBSplittingPass> X("bbsplit", "Ensures that all basic blocks contain at most one call or one conditional branch.",
-                             false /* Only looks at CFG */,
-                             false /* Analysis Pass */);
+static llvm::RegisterPass<polytracker::BBSplittingPass>
+    X("bbsplit",
+      "Ensures that all basic blocks contain at most one call or one "
+      "conditional branch.",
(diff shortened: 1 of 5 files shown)


ESultanik:
The reason for this transformation is to be able to perfectly temporally reconstruct the order of canonical byte accesses in each BB. Without it, it will be ambiguous for any BB that has a function call whether the taints were before or after the call. With it, we know that any taints occurred before the call.

polytracker/src/dfsan_pass/BBSplittingPass.cpp
```diff
@@ -0,0 +1,93 @@
...
+        }
+      }
+      // We need to split this BB into a new one after the call
+      modified = true;
+      bool includeFunctionName =
+          (fname != "llvm.dbg.declare" && fname != "__assert_fail");
+      // Don't bother logging these common functions, but still split for them
+      auto bb = next->getParent();
+      if (fname.length() == 0 || bb->hasName() || includeFunctionName) {
+        std::cout << "Splitting basic block";
```
carsonharmon, line 62: Let's block all the printing to stdout with some #ifdef debug guards.

polytracker/src/dfsan_pass/BBSplittingPass.cpp
```diff
@@ -0,0 +1,93 @@
...
+        if (bb->hasName()) {
+          std::cout << " " << bb->getName().data();
```
carsonharmon, line 64: same

polytracker/src/dfsan_pass/BBSplittingPass.cpp
```diff
@@ -0,0 +1,93 @@
...
+        }
+        if (includeFunctionName) {
+          std::cout << " after call to " << fname;
```
carsonharmon, line 67: same

polytracker/src/dfsan_pass/BBSplittingPass.cpp
```diff
@@ -0,0 +1,93 @@
...
+        }
+        std::cout << std::endl;
```
carsonharmon, line 69: same

polytracker/src/dfsan_pass/BBSplittingPass.cpp
```diff
@@ -0,0 +1,93 @@
...
+          fname = v->getName().data();
+        }
+      }
+      // We need to split this BB into a new one after the call
+      modified = true;
+      bool includeFunctionName =
+          (fname != "llvm.dbg.declare" && fname != "__assert_fail");
+      // Don't bother logging these common functions, but still split for them
+      auto bb = next->getParent();
+      if (fname.length() == 0 || bb->hasName() || includeFunctionName) {
```
carsonharmon, line 61: I'm curious about this check, what does fname.length() == 0 mean in this case? That some function is anonymous? I don't know how that fits into the splitting criteria but would be interested to know

polytracker/src/dfsan_pass/BBSplittingPass.cpp
```diff
@@ -0,0 +1,93 @@
...
+        if (function->hasName()) {
+          fname = function->getName().data();
+        }
+      } else if (auto v = call->getCalledValue()->stripPointerCasts()) {
+        if (v->hasName()) {
+          fname = v->getName().data();
+        }
+      }
+      // We need to split this BB into a new one after the call
+      modified = true;
```
carsonharmon, line 56: I don't see if this variable is used anywhere else besides the return, but I might be blind. Maybe we can just return True for this branch where we split, and false for the other one (which you already do)

polytracker/src/dfsan_pass/BBSplittingPass.cpp
```diff
@@ -0,0 +1,93 @@
...
+          std::cout << " " << bb->getName().data();
+        }
+        if (includeFunctionName) {
+          std::cout << " after call to " << fname;
+        }
+        std::cout << std::endl;
+      }
+      bb->splitBasicBlock(next);
+    }
+    return modified;
```
carsonharmon, line 73: Hmm so it looks like we always return in the middle of this loop body, meaning that after the first instruction we return if it was modified or not. Should the be outside of the loop?

polytracker/src/dfsan_pass/DataFlowSanitizer.cpp
```diff
@@ -968,12 +969,19 @@ bool DataFlowSanitizer::runOnModule(Module &M) {
...
     if (!i || i->isDeclaration())
       continue;
 
     removeUnreachableBlocks(*i);
 
+    // Split the function's basic blocks such that each BB has at most one
+    // function call and one conditional branch. This means that all of the
+    // calls in the instrumented code will be followed by an unconditional
+    // branch.
+    bbSplitter.runOnFunction(*i);
```
carsonharmon, line 983: Some more thoughts on this, even though I know it slows down the progress a bit. 

Since we are using opt to load/run passes with polybuild, I think we can write some simple testcases in the form of analysis pass. So we take some code, we generate the bitcode, we run basic block splitter pass on it, 

then our testcase takes that .bc and verifies that each block only has one call or branch statement in it. It would make me feel more confident if we had a testcase or two, I will work on this while you are away on PTO so we can  be closer to merging

polytracker/src/dfsan_pass/BBSplittingPass.cpp
ESultanik: I encountered some functions that had no name. I didn't investigate what caused it, but I presume it was either an anonymous function or a function that is guaranteed to be inlined.

polytracker/src/dfsan_pass/BBSplittingPass.cpp
ESultanik: So, uh, I apologize, but it looks like I fixed this code in the `mimid` branch and didn't push it back to this branch 😫 Sorry for the confusion, and thank you for reviewing this, but I think I should just close this PR and we should focus on the version in PR #93

polytracker/src/dfsan_pass/DataFlowSanitizer.cpp
ESultanik: That's a good idea. I am going to end up closing this PR, but I will create an issue to track this.

PR #95, opened 2020-09-10
Canonical byte lookup bugfixes and datastructure optimization

ESultanik:
@carsonharmon this is a small one that fixes the canonical byte lookup for grammar extraction. I changed some of the instrumentation datastructures and how we iterate over them, so can you please take a quick look to make sure it all seems sane? Feel free to merge this yourself if/when it looks good.

### DIFF
--- a/polytracker/src/dfsan_rt/dfsan/dfsan_log_mgmt.cpp
+++ b/polytracker/src/dfsan_rt/dfsan/dfsan_log_mgmt.cpp
@@ -286,13 +286,13 @@ void taintManager::addJsonRuntimeTrace() {
           std::transform(taints.begin(), taints.end(),
                          std::back_inserter(byteOffsets),
                          [&mapping](dfsan_label d) {
-            for (const auto& pair : mapping) {
-              if (pair.first == d) {
-                return pair.second;
-              }
-            }
-            return -1;
-          });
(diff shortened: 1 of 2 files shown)


polytracker/src/dfsan_rt/dfsan/dfsan_log_mgmt.cpp
```diff
@@ -280,13 +281,19 @@ void taintManager::addJsonRuntimeTrace() {
         }
         const auto& taints = trace.taints(bb);
         if (!taints.empty()) {
-          // dfsan_labels are 1-indexed
-          std::vector<dfsan_label> zeroIndexed;
-          zeroIndexed.reserve(taints.size());
+          std::vector<int> byteOffsets;
```
carsonharmon, line 284: Question: Does the `taints` set only have taint labels that are canonical?

carsonharmon (changes requested):

polytracker/src/dfsan_rt/dfsan/dfsan_log_mgmt.cpp
ESultanik: Correct

carsonharmon (approved):

PR #102, opened 2020-10-20
Carson/new trace

carsonharmon:
This PR adds a few new things

1. This removes dfsan_log_mgmt, and adds a new directory polytracker/src 
2. Polytracker/src has -O3 and -Wall -Werror, bumps the standard to C++17 
3. It removes the use of the taint manager, and instead replaces it with some functions
4. Some C++ code improvements 

- Use pass by reference, really just avoid not making extra copies
- Remove unneeded object creation
- Use attribute [nodiscard] to have the compiler warn when we aren't using a value properly 
- Use lock_guards to remove manual locking/unlocking 
- Use thread local to avoid needing a shared mono lock all the time 

5. This adds a new function dfsan_trace_inst_fn which currently does nothing 
6. This PR comments out the optimization to not output some functions in the JSON since it crashes
7. Removes the roaring libs, we stopped using them forever ago anyway

Probably more stuff, plz review

Overall this should remove bloat, and make the codebase easier to maintain/upgrade/modify since everything is in src/polytracker.


ESultanik (approved):
I'm not sure how I feel about prepending the artifact files with `#_` for each thread, but I can't think of a better way using JSON. I guess we should just switch to sqlite. But this is fine for now.

PR #103, opened 2020-10-27
Carson/config file support

carsonharmon:
This needs to be merged after #102 

This PR adds support for config files 

1. First, tries detecting if a config file in the current directory exists ("./polytracker_config.json") 
2. If not, tries detecting if a config file at "~/.config/polytracker/polytracker_config.json" exists 
3. If a config file is found, it parses it and fills in the appropriate settings 
4. the environment variables are then parsed, the env settings will overwrite whatever is in the config. 
5. for settings not touched by the config or env, defaults are set or an error is thrown. 

Also adds a test in test_polytracker

So the priorities for settings are ENV > Config > Defaults

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,10 +135,34 @@ POLYSTART: Start offset to track
 
 POLYEND: End offset to track
 
-POLYOUTPUT: Provides a a path/prefix to output polytracker information too
+POLYOUTPUT: Provides a path/prefix to output polytracker information too
+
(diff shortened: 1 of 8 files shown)


ESultanik (changes requested):
The code looks good, but we should add usage info to the `README.md`.

ESultanik (approved):

PR #106, opened 2020-12-10
(WIP) Carson/sqlite

carsonharmon:
This is not ready to merge, but this PR adds a few things 

- [x] Replaces the use of JSON and taint forest output by inserting into a sqlite3 database 
- [x] Updates polytracker to use the databases instead of json/forests
- [x] Updates the tests to do the same 
- [x] Adds some new polytracker commands for listing input ID info 
- [x] Adds some documentation and a database schema diagram
- [x] Also, I think I made some instrumentation changes this PR to swap from using function names everywhere to indexes provided by the trace instrumentation 
- [x] Some more misc cleaning/improvements


ESultanik:
This PR is subsumed by #6346.

PR #6360, opened 2021-04-19
Various database schema and access API changes for a speedup

ESultanik:
`TaintAccess`es are no longer events and each have their own sequential global UID. Basic block entry counts are calculated at runtime rather than in post-processing.

### DIFF
--- a/polytracker/grammars.py
+++ b/polytracker/grammars.py
@@ -622,6 +622,9 @@ def trace_to_grammar(trace: ProgramTrace) -> Grammar:
     grammar = Grammar()
 
     entrypoint = trace.entrypoint
+    if entrypoint is None:
+        grammar.start = Production(grammar, "<START>")
+        return grammar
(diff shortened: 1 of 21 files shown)


polytracker/grammars.py
```diff
@@ -612,98 +612,95 @@ def __str__(self):
...
+                    if called_function is not None:
+                        func_stack.append(called_function)
+                        sub_productions.append(production_name(called_function))
+                        ret = called_function.function_return
+                        if ret is not None:
+                            returning_to = ret.returning_to
+                            if returning_to is not None:
+                                sub_productions.append(f"<{returning_to!s}>")
+                            else:
+                                # TODO: Print warning
```
carsonharmon, line 682: Did you want to add these warnings before merge?

ESultanik: A lot of that code gets removed in the forthcoming `early_parser` branch (see #6361), so we'll address it there.

PR #6373, opened 2021-04-26
Remove  references from logging which prevent TCO

carsonharmon:
@ESultanik I removed some references here, I think should be good to go. 

Doing this before doing the experiments.


ESultanik (approved):
Looks good to me. It's ready to merge if/when all of the tests pass.

Issue #6354, opened 2021-04-08
Track all inputs by default


ESultanik:
Implementing this should block on the completion of #6353.
- [ ] Do not require `POLYPATH`. If omitted, track _all_ inputs encountered.
- [ ] Allow `POLYPATH` to contain a colon-delimited list of input files to track.

PR #6374, opened 2021-04-27
If no polypath is specified, always taint

carsonharmon:
This makes us always taint when POLYPATH is not specified.

### DIFF
--- a/polytracker/src/polytracker/main.cpp
+++ b/polytracker/src/polytracker/main.cpp
@@ -16,6 +16,7 @@
 #include <sys/mman.h>
 #include <sys/stat.h>
 #include <unistd.h>
+#include <unordered_set>
 
 using json = nlohmann::json;
 
@@ -34,9 +35,7 @@ bool polytracker_trace_func = false;
 bool polytracker_save_input_file = true;
 decay_val taint_node_ttl = 0;
 // If this is empty, taint everything.
-std::string target_file = "";
-bool track_all_taint = false;
-
+std::unordered_set<std::string> target_sources;
 char *forest_mem;
 std::atomic_bool done = ATOMIC_VAR_INIT(false);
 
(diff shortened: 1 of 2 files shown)


polytracker/src/polytracker/main.cpp
```diff
@@ -24,16 +24,19 @@ using json = nlohmann::json;
...
 bool polytracker_trace = false;
 bool polytracker_trace_func = false;
 /**
  * Whether or not to save the input files to the output database
  */
 bool polytracker_save_input_file = true;
 decay_val taint_node_ttl = 0;
+// If this is empty, taint everything.
 std::string target_file = "";
+bool track_all_taint = false;
```
ESultanik, line 38: Instead of having a bool to toggle between either tracking a single file or tracking all taints, I think we should instead get rid of this and also get rid of `target_file`. Instead, create a `std::unordered_set<std::string> target_files` containing all of the whitelisted files, generated by splitting `POLYPATH` delimited by colons. If `POLYPATH` is empty or undefined, then `target_files` will be empty. If `target_files` is empty, then implicitly default to tracking all files. See #6354.

ESultanik (approved):
This looks good, but it only partially satisfies #6354.

polytracker/src/polytracker/main.cpp
carsonharmon: Done! There is now a `target_sources` set and some logic to parse strings in the form of `source1;source2` etc. If nothing is specified, everything is tainted.

polytracker/src/polytracker/main.cpp
```diff
@@ -280,8 +287,10 @@ void polytracker_start() {
   polytracker_print_settings();
   output_db = db_init(polytracker_db_name);
   // Store new file
-  input_id = storeNewInput(output_db, target_file, byte_start, byte_end,
-                           polytracker_trace);
+  for (auto target_source : target_sources) {
+    input_id = storeNewInput(output_db, target_source, byte_start, byte_end,
```
ESultanik, line 291: What happens if `target_sources` is empty? Do we call `storeNewInput` if/when we encounter a new input?

ESultanik (changes requested):
I have a bunch of suggestions I typed into GitHub. NOTE: I did not actually test compiling them! 😅

ESultanik (approved):

PR #6382, opened 2021-07-22
XRay Instrumentation + Some Changes

carsonharmon:
This branch adds some new features and performance improvements. 

1. Its now possible to build polytracker and targets with LLVM Xray instrumentation. For polytracker its a cmake setting, for targets its a setting in Polybuild.
 
2. We've brought back end of function instrumentation for logging exits, this fixes the mismatch issue with stack frames. We still have post call instrumentation, which is for exceptions/setjmp/longjmp/other stack jumping etc 

3. We now store the function call graph into the program now, which we can use later to replace some of the information we generate at runtime 

4. Instrumentation calls polytracker_start at the beginning of main 

5. Some additional functions were added to ignore lists, mostly related to program start up 

6. There were many if statements that constantly checked for "done" to see if polytracker was/is still running. The Xray instrumentation showed that we spent a lot of time in our logging functions, so these additional checks were causing issues. By adding additional functions to the ignore list and starting at the entry point (main), we can remove these. 

7. XRay instrumentation also showed that we were spending a lot of time in our storing functions. We were doing redundant computation by repeatedly preparing the statements for every execution. After executing a sql statement we now reset it so it can be reused again.  

8. Adds some additional scripts for ignoring specific parts of libcxx. 

9. Adds example docker file for xpdf


polytracker/cfg.py
```diff
@@ -70,14 +65,7 @@ def ancestors(self, node: N) -> OrderedSet[N]:
...
         return OrderedSet(
-            *(
-                x
-                for _, x in sorted(
-                    (d, n)
-                    for n, d in nx.shortest_path_length(self, target=node).items()
-                    if n is not node
-                )
-            )
+            *(x for _, x in sorted((d, n) for n, d in nx.shortest_path_length(self, target=node).items() if n is not node))
```
ESultanik, line 68: I know this is "more Pythonic", but I think the old version was more readable

ESultanik (approved):

PR #6388, opened 2021-09-23
Support arm64, more recent libc and fix a bug in a testcase

hbrodin:


hbrodin:
I don't expect this to work until the updated polytracker-llvm image is pushed to Docker Hub

ESultanik (approved):

polytracker/src/taint_sources/taint_sources.cpp
```diff
@@ -445,10 +445,10 @@ EXT_C_FUNC int __dfsw_munmap(void *addr, size_t length, dfsan_label addr_label,
   return ret;
 }
 
-EXT_C_FUNC int __dfsw__IO_putc(int __c, FILE *__fp, dfsan_label c_label,
```
carsonharmon, line 448: There is a conflict related to IO_putc in one of the integration tests, which are running on x86_64. 

I could look into it if you want since I have an x86 machine? 

I think MuPDF uses io_putc on x86_64, or maybe not with an updated polytracker-llvm version?

It's not totally clear why there is an issue here.

carsonharmon (changes requested):

polytracker/src/taint_sources/taint_sources.cpp
hbrodin: This should go away with the updated libc-version from ubuntu:focal, which we get in the updated polytracker-llvm image. Maybe I should have made some effort here to be able to handle both cases. I will have a look tomorrow to see if I"m able to figure something out.

hbrodin:
Other PRs have solved the contents of this.

PR #6389, opened 2021-09-23
Carson/metadata tagging

carsonharmon:
This PR adds a few things

1. Metadata tagging. This is a new optional path for debugging instrumentation/examining instrumentation overhead. It marks the original instructions in the program, this could be useful for other things later. It does not add any function calls but works by adding metadata to the bitcode file, so no performance concerns. 

2. This PR stores the target program into the sqlite database at runtime. 

3. Extracts the bitcode from the program, and stores compile time artifacts into the database. In this PR, it *removes* storeFunc from the runtime hot path. In a follow up PR, we can remove storeBlock as well. 

There are a few unholy things I had to do to get this to work, I'll lay out the secrets here. 

A. In order to extract and parse the bitcode from the program, I link LLVM to the polytracker runtime. Normally this would be fine, but Polytracker is built with it's own libcxx, which is different from the one LLVM is built with. I attempted to build LLVM with its own libcxx, but failed multiple times. Without doing it though, you would get errors like this:
 
`/polytracker/build/../polytracker/src/polytracker/output.cpp:352: undefined reference to `llvm::parseIRFile(llvm::StringRef, llvm::SMDiagnostic&, llvm::LLVMContext&, llvm::function_ref<llvm::Optional<std::__p::basic_string<char, std::__p::char_traits<char>, std::__p::allocator<char> > > (llvm::StringRef)>)'`

Notice the `std::__p`, this indicates that the string type is located in polytracker's internal ABI. This is desired behavior, as it prevents overlap when analyzing C/C++ programs. 

The way I solved this issue without rebuilding/changing polytracker-llvm is by using the LLVM C-API for functions that directly touch the ABI, and then using a previously unknown to me LLVM function called `unwrap` I can convert the C types to C++ types, and it works!

B. After successfully building the target with libPolytracker + llvm-link libs, there were issues extracting the bitcode. 

B1. The global structure we wanted to parse and store into the database was gone 
B2. Linking errors upon extraction from llvm-link. 

My intuition for B1 was that the compiler was optimizing away the unused global. The compiler is unable to optimize away any data or instruction when it does not have clear visibility in how its used, so I added an additional function in `main` during start up that takes our global as a parameter, and does nothing. This has no performance overhead as its a void function that is only called once, and it guarantees that the compiler cant delete the data we care about. 

B2 I had to create a new wrapper called poly-link. The issue is the same as we had when originally building programs, that the ABI wont let you add the std::__p prefix to everything during compile time options, so there are symbol conflicts. These symbols wont be instrumented anyway, so all we need to do is pass the `--only-needed` flag and it will lazily link symbols. 

This fixes all the issues. Tests pass locally.


polytracker/database.py
```diff
@@ -638,13 +642,66 @@ def basic_blocks(self) -> Iterator[BasicBlockEntry]:
...
+
+    def func_name(self, id: int) -> str:
+        if self._func_map is None:
+            self._func_map = self.extract_func_map()
+        if id not in self._func_map:
+            raise ValueError(f"Function id: {id} not in func map!")
+        return self._func_map[id]
+
+    @property
+    def block_map(self):
```
ESultanik, line 677: Doesn't this need to be implemented?

polytracker/database.py
```diff
@@ -638,13 +642,66 @@ def basic_blocks(self) -> Iterator[BasicBlockEntry]:
...
+        pass
+
+    @property
+    def nomina_map(self):
```
ESultanik, line 681: Doesn't this need to be implemented?

polytracker/scripts/poly-link.py
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python3
+import sys
+import subprocess
+
+args = [x for x in sys.argv[1:]]
```
ESultanik, line 5:
```suggestion
args = list(sys.argv[1:])
```

polytracker/scripts/poly-link.py
```diff
@@ -0,0 +1,7 @@
...
+final = ["llvm-link", "--only-needed"] + args
+subprocess.check_call(final)
```
ESultanik, line 7: PEP compliance (newline at end of file):
```suggestion
subprocess.check_call(final)

```

polytracker/scripts/polybuild.py
```diff
@@ -149,14 +155,18 @@ def instrument_bitcode(bitcode_file: Path, output_bc: Path,
     subprocess.check_call(opt_command)
     if ignore_lists is None:
         ignore_lists = []
+    opt_command = ["opt", "-enable-new-pm=0", "-load", str(META_PASS_PATH), "-meta", str(bitcode_file), "-o", str(bitcode_file)]
+    ret = subprocess.call(opt_command)
+    if ret != 0:
+        print("Metadata pass failed! Exiting!")
```
ESultanik, line 161: Better to also print the command itself so the user can debug it:
```suggestion
        print(f"Metadata pass exited with code {ret}:\n{' '.join(opt_command)}")
```

polytracker/scripts/polybuild.py
```diff
@@ -149,14 +155,18 @@ def instrument_bitcode(bitcode_file: Path, output_bc: Path,
...
     opt_command = ["opt", "-enable-new-pm=0", "-load",
                    str(POLY_PASS_PATH), "-ptrack", f"-ignore-list={POLY_ABI_LIST_PATH!s}"]
     if file_id is not None:
         opt_command.append(f"-file-id={file_id}")
     for item in ignore_lists:
         opt_command.append(f"-ignore-list={ABI_PATH}/{item}")
     opt_command += [str(bitcode_file), "-o", str(output_bc)]
-    print(opt_command)
     ret = subprocess.call(opt_command)
     assert ret == 0
```
ESultanik, line 171:
```suggestion
    if ret != 0:
        print(f"PolyTracker pass exited with code {ret}:\n{' '.join(opt_command)}")
```

polytracker/src/CMakeLists.txt
```diff
@@ -8,20 +8,33 @@ if (LLVM_FOUND)
...
+  INTERFACE_LINK_DIRECTORIES "${LLVM_LIBRARY_DIRS}"
+  INTERFACE_COMPILE_DEFINITIONS "LLVM_VERSION_MAJOR=${LLVM_VERSION_MAJOR};LLVM_VERSION_MINOR=${LLVM_VERSION_MINOR}"
+  # INTERFACE_LINK_OPTIONS "-Wl,-znodelete"
+  )
+
 set(POLY_DIR "polytracker")
 set(TAINT_DIR "taint_sources")
 
-set(CMAKE_CXX_STANDARD 17)
+# set(CMAKE_CXX_STANDARD 17)
```
ESultanik, line 28: Why was this commented out?

polytracker/src/passes/metadata_pass.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+namespace polymeta {
+
+static llvm::Constant *create_str(llvm::Module &mod, std::string &str) {
+  auto arr_ty = llvm::ArrayType::get(
+      llvm::IntegerType::getInt8Ty(mod.getContext()), str.size() + 1);
+
+  auto int8_ty = llvm::IntegerType::getInt8Ty(mod.getContext());
+
+  // TODO (Carson) this feels hacky
```
ESultanik, line 46: I'm not sure of a cleaner way to do it…

polytracker/src/polytracker/output.cpp
```diff
@@ -8,12 +8,25 @@
...
+#include <llvm-c/BitReader.h>
+#include <llvm-c/BitWriter.h>
+#include <llvm-c/Core.h>
+#include <llvm-c/Support.h>
+#include <llvm/IR/Constants.h>
+#include <llvm/IR/LLVMContext.h>
+#include <llvm/IR/Module.h>
+#include <llvm/Support/SourceMgr.h>
+
+#define POLY_TEMP_FILE "/tmp/polytracker_temp.binary"
```
ESultanik, line 28: This is probably fine for now, but we might get some weird bugs if/when a system ever doesn't have a writable `/tmp` directory. It'd probably be better to use `mktemp`.

polytracker/src/polytracker/polytracker.cpp
```diff
@@ -99,15 +101,17 @@ extern "C" void __polytracker_log_bb_entry(char *name, uint32_t findex,
 extern "C" atomic_dfsan_label *
 __polytracker_union_table(const dfsan_label &l1, const dfsan_label &l2) {
   // if (LIKELY(!done)) {
```
ESultanik, line 103:
```suggestion
```

polytracker/src/polytracker/polytracker.cpp
```diff
@@ -99,15 +101,17 @@ extern "C" void __polytracker_log_bb_entry(char *name, uint32_t findex,
...
-  try {
-    return getUnionEntry(l1, l2);
-  } catch (std::exception &e) {
-    return nullptr;
-  }
+  // try {
```
ESultanik, line 104:
```suggestion
```

polytracker/src/polytracker/polytracker.cpp
```diff
@@ -99,15 +101,17 @@ extern "C" void __polytracker_log_bb_entry(char *name, uint32_t findex,
...
+  return getUnionEntry(l1, l2);
+  //} catch (std::exception &e) {
```
ESultanik, line 106:
```suggestion
```

polytracker/src/polytracker/polytracker.cpp
```diff
@@ -99,15 +101,17 @@ extern "C" void __polytracker_log_bb_entry(char *name, uint32_t findex,
...
+  // return nullptr;
```
ESultanik, line 107:
```suggestion
```

polytracker/src/polytracker/polytracker.cpp
```diff
@@ -99,15 +101,17 @@ extern "C" void __polytracker_log_bb_entry(char *name, uint32_t findex,
...
+  //}
```
ESultanik, line 108:
```suggestion
```

polytracker/src/polytracker/polytracker.cpp
```diff
@@ -99,15 +101,17 @@ extern "C" void __polytracker_log_bb_entry(char *name, uint32_t findex,
...
   //}
```
ESultanik, line 109:
```suggestion
```

polytracker/src/polytracker/polytracker.cpp
```diff
@@ -99,15 +101,17 @@ extern "C" void __polytracker_log_bb_entry(char *name, uint32_t findex,
...
   // return nullptr;
```
ESultanik, line 110:
```suggestion
```

ESultanik (changes requested):

ESultanik (approved):

PR #6408, opened 2021-11-02
Remove old ABI lists

ESultanik:
Only keep the ABI lists that we actually use now.


ESultanik:
This PR is now subsumed by #6410.

PR #6410, opened 2021-11-02
Detect "cavities" in a file

ESultanik:
This adds an instrumentation-time option `--no-control-flow-tracking` that does not add any control-flow tracking instrumentation. It also adds support for tracking taints that affect conditional branches. It also adds a Python command for detecting "file cavities": portions of the input file that neither influence control flow nor bytes that are written to an output file.

```
$ polytracker cavities polytracker.db
```


hbrodin:
Would it make sense to somehow unify the code used to build mupdf? We now have one set of commands in Dockerfile-mupdf.demo and another in build_in_docker.

polytracker/custom_abi/dfsan_abilist.txt
```diff
@@ -183,7 +185,7 @@ fun:fputs=discard
 fun:fputs_unlocked=discard
 fun:fseek=discard
 fun:ftell=discard
-fun:fwrite=discard
+#fun:fwrite=discard
```
hbrodin, line 188: Seems redundant? Have fwrite=custom below.

polytracker/custom_abi/polytracker_abilist.txt
```diff
@@ -183,7 +185,7 @@ fun:fputs=discard
 fun:fputs_unlocked=discard
 fun:fseek=discard
 fun:ftell=discard
-fun:fwrite=discard
+#fun:fwrite=discard
```
hbrodin, line 188: Redundant? (And not used due to '#').

polytracker/database.py
```diff
@@ -875,31 +885,40 @@ class DBTaintOutput(Base, TaintOutput):  # type: ignore
     __tablename__ = "output_taint"
     input_id = Column(Integer, ForeignKey("input.id"))
     offset = Column(BigInteger)
-    label = Column(BigInteger)
+    label = Column(BigInteger, ForeignKey("taint_forest.label"))
+    taint_forest_node: "DBTaintForestNode" = relationship("DBTaintForestNode", foreign_keys=[label],
```
hbrodin, line 889: Why quoted "DBTaintForestNode"?

polytracker/database.py
```diff
@@ -942,7 +961,7 @@ def parent_two(self) -> Optional["DBTaintForestNode"]:
         return self._parent_two
 
     @staticmethod
-    def taints(labels: Iterable["DBTaintForestNode"]) -> Taints:
+    def get_taints(labels: Iterable["DBTaintForestNode"]) -> Taints:
```
hbrodin, line 964: I believe this should be reflected in the examples in README.md

polytracker/include/polytracker/main.h
```diff
@@ -5,7 +5,7 @@ struct func_mapping;
 struct block_mapping;
 
 void polytracker_start(func_mapping const *globals, uint64_t globals_count,
-                       block_mapping const *block_map,
-                       uint64_t block_map_count);
+                       block_mapping const *block_map, uint64_t block_map_count,
+                       bool control_flow_tracing);
```
hbrodin, line 9: minor: Called control_flock_tracking elsewhere

polytracker/include/polytracker/polytracker.h
```diff
@@ -38,8 +38,8 @@ struct block_mapping {
 
 void polytracker_end();
 void polytracker_start(func_mapping const *globals, uint64_t globals_count,
-                       block_mapping const *block_map,
-                       uint64_t block_map_count);
+                       block_mapping const *block_map, uint64_t block_map_count,
+                       bool control_flow_tracing);
```
hbrodin, line 42: (sizeof(T)*8-1)/7+1

polytracker/mapping.py
```diff
@@ -0,0 +1,146 @@
...
+        return ret
+
+    def written_input_bytes(self) -> Iterator[ByteOffset]:
+        """Yields all of the input byte offsets from input files that are written to an output file"""
+        # forest = self.trace.taint_forest
+        # yield from self.trace.taints(itertools.chain(*(
+        #     forest[output.label] for output in tqdm(self.trace.output_taints, unit=" output taints", leave=False)
+        #     if output.label > 0
+        # )))
+        # return
```
hbrodin, line 55: remove?

polytracker/src/passes/polytracker_pass.cpp
```diff
@@ -543,82 +572,105 @@ bool PolytrackerPass::runOnModule(llvm::Module &mod) {
...
-  for (auto &func : mod) {
-    // Ignore if its in our ignore list
-    if (func.hasName()) {
-      std::string fname = func.getName().str();
-      if (ignore_funcs.find(fname) != ignore_funcs.end()) {
-        continue;
+  if (no_control_flow_tracking) {
+    std::cout << "Omitting PolyTracker control flow instrumentation."
+              << std::endl;
+    // We still want to visit comparison functions because they are useful
```
hbrodin, line 578: minor, for clarity: comparison instructions?

polytracker/tracing.py
```diff
@@ -1077,6 +1095,51 @@ def taint_forest(self) -> TaintForest:
...
+        ) as t:
+            while node_stack:
+                node = node_stack.pop()
+                t.update(node.label)
+                if node.parent_one is None:
+                    assert node.parent_two is None
+                    taints.add(self.file_offset(node))
+                else:
+                    parent1, parent2 = node.parent_one, node.parent_two
+                    assert parent1 is not None and parent2 is not None
```
hbrodin, line 1131: Why is it required for a node to have two parents? Is this because node labels are reused if it is only involved in an unary expression? Maybe a minor comment here about the reasoning?

polytracker/include/polytracker/polytracker.h
hbrodin, line 42:
```suggestion
                       bool control_flow_tracking);
```

polytracker/include/polytracker/main.h
hbrodin, line 9:
```suggestion
                       bool control_flow_tracking);
```

hbrodin (approved):
Overall it looks very good to me! There are some suggestions that you might consider. Mainly to improve ease of understanding the code.
Regarding how we build the mutool, if you think we should unify it we could create a separate issue and still merge this.

ESultanik:
> Would it make sense to somehow unify the code used to build mupdf? We now have one set of commands in Dockerfile-mupdf.demo and another in build_in_docker.

Yes! I think that can wait for another PR, but it would be great to clean up the examples and merge them with this new build process.

polytracker/database.py
ESultanik: It's one of the quirks of Python type-hinting. We don't define the `DBTaintForestNode` class until later in the file, so we can't reference the type yet. Referencing the type explicitly here would cause an error immediately when the module is loaded. The workaround is to sort-of forward declare the type by putting it in a string.

polytracker/database.py
ESultanik: Ugh, it's complicated. This change should not affect the public API, because a `taints` function still exists in the superclass, so the examples in the docs should still work. I renamed the static version to `get_taints` in order to avoid the name conflict with the version used in the documentation that is not static.

polytracker/mapping.py
ESultanik:
```suggestion
```

polytracker/src/passes/polytracker_pass.cpp
ESultanik: Good catch!
```suggestion
    // We still want to visit comparison instructions because they are useful
```

polytracker/tracing.py
ESultanik: That is correct.
```suggestion
                    // a node will always have either zero or two parents.
                    // labels that are reused will reuse their associated nodes.
                    // all other nodes are unions.
                    assert parent1 is not None and parent2 is not None
```

polytracker/database.py
hbrodin: Ahh... Of course. It makes sense. I wasn't aware of this.

PR #6412, opened 2022-01-21
Taint dag

hbrodin:
Just opening the pull request to be able to discuss the code.


polytracker/dumptdag.py
```diff
@@ -0,0 +1,163 @@
...
+from io import SEEK_SET
+from mmap import mmap, PROT_READ
+from pathlib import Path
+from typing import Iterable, Tuple, Union
+import sys
+
+
+#TODO (hbrodin): Completely unchecked values. Only parse files from trusted sources...
+
+# This needs to be kept in sync with implementation in encoding.cpp
```
surovic, line 12: Let's bake this into the tdag file as a header or something of that sort.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM trailofbits/polytracker-llvm:f1385f079c2573f2b31834354310e7265a9bb4b3
+FROM trailofbits/polytracker-llvm:latest
```
surovic, line 1: We need to remember to pin this again to a specific image once we're done with the PR.

polytracker/include/taintdag/fdmapping.hpp
```diff
@@ -0,0 +1,133 @@
...
+       : begin_{begin}, end_{end} {
+      }
+
+      size_t get_size() const {
+        std::unique_lock l{m_};
+        return offset_;
+      }
+
+      // The only reason this could fail is if there is no space left (either storage or source_index)
+      std::optional<source_index_t> add_mapping(int fd, std::string_view name, std::optional<taint_range_t> preallocated_labels= {}) {
```
hbrodin, line 46: Consider keeping the mappings separate from the filenames, to be able to directly index into the mappings. Add a name offset field into each header where to locate the filename.

polytracker/taint_dag.py
```diff
@@ -0,0 +1,360 @@
...
+            yield (path, fdhdr)
+            offset += len(path)
+
+    @property
+    def label_count(self):
+        return int(self.header.tdag_mapping_size / sizeof(c_uint64))
+
+    def read_node(self, label: int) -> int:
+        if label in self.raw_nodes:
+            return self.raw_nodes[int]
```
hbrodin, line 140: Any speedup from this caching? I would assume it would make things slower since the results is more or less just a read from the mmap-region? But, as is common with optimizations it is hard to guess...

surovic: Probably not to be honest. I'll see if it makes any difference once I have cavity detection working and remove if not.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM trailofbits/polytracker-llvm:f1385f079c2573f2b31834354310e7265a9bb4b3
+FROM trailofbits/polytracker-llvm:16daa680dcff0dc86ebf6ae81f5382859695cb35
```
facutuesca, line 1: Can you remove `sqlite3` and `libsqlite3-dev` from the installed packages, or are they still needed?

hbrodin: Yes! Good point! That should work. Also, libjpeg-dev shouldn't be needed. Maybe others.

PR #6415, opened 2022-01-28
Instrumenting switch cases to detect taint that influence control flow

hbrodin:
Previously, only branch instructions was only considered to affect control flow. This PR instruments switch instructions as well. Switch instructions do also affect control flow.

### DIFF
--- a/polytracker/src/passes/polytracker_pass.cpp
+++ b/polytracker/src/passes/polytracker_pass.cpp
@@ -138,7 +138,6 @@ void PolytrackerPass::visitBranchInst(llvm::BranchInst &BI) {
 void PolytrackerPass::visitSwitchInst(llvm::SwitchInst &SI) {
   // TODO (hbrodin): Is a guard needed? E.g. op_check as in visitBranchInst?
   llvm::IRBuilder<> IRB(&SI);
-  llvm::LLVMContext &context = mod->getContext();
   llvm::Value *int_val = IRB.CreateSExtOrTrunc(SI.getCondition(), shadow_type);
   CallInst *Call = IRB.CreateCall(conditional_branch_log, {int_val});
 }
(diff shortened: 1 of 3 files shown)


polytracker/src/passes/polytracker_pass.cpp
```diff
@@ -132,6 +135,14 @@ void PolytrackerPass::visitBranchInst(llvm::BranchInst &BI) {
   }
 }
 
+void PolytrackerPass::visitSwitchInst(llvm::SwitchInst &SI) {
+  // TODO (hbrodin): Is a guard needed? E.g. op_check as in visitBranchInst?
+  llvm::IRBuilder<> IRB(&SI);
+  llvm::LLVMContext &context = mod->getContext();
```
surovic, line 141: This seems redundant?

polytracker/src/passes/polytracker_pass.cpp
```diff
@@ -589,6 +600,8 @@ bool PolytrackerPass::runOnModule(llvm::Module &mod) {
         for (auto &inst : bb) {
           if (auto *BI = llvm::dyn_cast<llvm::BranchInst>(&inst)) {
             visitBranchInst(*BI);
+          } else if (auto *SI = llvm::dyn_cast<llvm::SwitchInst>(&inst)) {
```
surovic, line 603: We can avoid the whole `for(auto &inst : bb){...}` by using `visit(bb->getTerminator())`.

surovic (approved):

polytracker/src/passes/polytracker_pass.cpp
hbrodin: Great idea! We can't do exactly as you write, but we don't need to loop. If we use the visitor we would also get e.g. the ret-instr visitation and that is not desired.

polytracker/src/passes/polytracker_pass.cpp
hbrodin: True

polytracker/src/passes/polytracker_pass.cpp
surovic: True, my bad! Does it matter when we visit branches and switches?

hbrodin: We would add instrumentation for the ret instruction and that typically instruments the instruction with a call to a method to log function exits.  I'll make a slight variation.

PR #6418, opened 2022-02-11
Parallel corpus runner modifications

surovic:


build_in_docker/file_cavity_detection.py
```diff
@@ -92,26 +105,29 @@ def file_cavity_detection(file: Path, output_dir: Path, timeout: int) -> str:
         if return_str == "":
             command = [
                 "python3",
-                str(SCRIPTDIR / "../polytracker/dumptdag.py"),
+                # str(SCRIPTDIR / "../polytracker/dumptdag.py"),
+                str(SCRIPTDIR / "cavities.py"),
```
hbrodin, line 109: I guess the `dumptdag.py` script is about to be retired :)

build_in_docker/file_cavity_detection.py
```diff
@@ -92,26 +105,29 @@ def file_cavity_detection(file: Path, output_dir: Path, timeout: int) -> str:
...
                 str(tmpd / TDAG),
-                f"/inputs/{filename}"
+                f"/inputs/{filename}",
```
hbrodin, line 111: why the additional "," ?

hbrodin (approved):
LGTM

build_in_docker/file_cavity_detection.py
surovic: Better leave it for a while longer. The polytracker one is still pretty slow :eyes:

PR #6421, opened 2022-02-17
Allow higher `typing_extensions` package versions and add stuff to .gitignore

surovic:


ESultanik (approved):

ESultanik:
Oops, just noticed that this is failing type checking in CI. We should probably fix that before merging.

ESultanik:
You can reproduce locally by running:
```console
mypy --ignore-missing-imports polytracker tests
```

PR #6425, opened 2022-02-22
Ahere to the POLYDB environment variable. Remove db-interaction.

hbrodin:
Stores the tdag file using env var POLYDB (default polytracker.tdag). Drops the sqlite db support. This is done to prevent the sqlite db and tdag file from overwriting each other.


surovic:
Probably time to jump the gun with this. LGTM.

surovic (approved):

PR #6430, opened 2022-02-28
 Formats all `.py` files via `black`

surovic:


ESultanik (approved):
Lints are now passing

PR #6431, opened 2022-03-01
Polytracker taint dag API tests

surovic:


polytracker/dumptdag.py
```diff
@@ -194,7 +194,7 @@ def dump_tdag(file: Path):
 # NOTE (hbrodin): Assumes source taint was preallocated
 
 
-def gen_source_taint_used(tdagpath: Path, sourcefile: Path) -> bytearray:
```
hbrodin, line 197: Why get rid of Path?

polytracker/taint_dag.py
```diff
@@ -385,22 +384,24 @@ def create_node(self, label: int) -> TDTaintForestNode:
             )
 
         elif isinstance(node, TDRangeNode):
-            for n in range(node.first, node.last):
+            curr: int = node.first
+            for n in range(node.first + 1, node.last):
```
hbrodin, line 388: When storing `last` and `first` in the tdag file, `last` is inclusive but the range-loop is exclusive of last element. Is this accounted for when parsing the range node? Otherwise we probably miss the last element?
Also, I am not all clear about why we start iteration from the second element in the range (`node.first + 1`)?

polytracker/taint_dag.py
```diff
@@ -324,8 +323,8 @@ def __repr__(self):
             f"label: {self.label} ; "
             f"input: {None if self.source is None else self.source.uid} ; "
             f"affected_control_flow: {self.affected_control_flow} ; "
-            f"parent_one: {self.parents[0]} ; "
-            f"parent_two: {self.parents[1]}"
+            f"parent_one: {self.parents[0] if self.parents else None} ; "
```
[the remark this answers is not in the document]
surovic: We do represent all taint forest nodes the same way. The bug was that when I tried to print a source node it tried to subscript a `None` and crashed.

polytracker/dumptdag.py
surovic: Reverted it back. Originally I had issues with comparisons and docker and other stuff, but managed to resolve it along the way.

polytracker/taint_dag.py
hbrodin: 👍

PR #6432, opened 2022-03-02
Maps files having same paths to the same source label range

hbrodin:
This essentially makes polytracker view files having same paths being equal.

### DIFF
--- a/polytracker/src/taintdag/fdmapping.cpp
+++ b/polytracker/src/taintdag/fdmapping.cpp
@@ -1,5 +1,29 @@
 #include "taintdag/fdmapping.hpp"
 
+#include <iterator>
+
+namespace {
+
+  // Helper structure that allows range-based for iteration
+  // on the entries in FDMapping.
+  template <typename It>
+  struct RetVal {
+    It b;
+    It e;
+    RetVal(It b, It e) : b{b}, e{e} {}
+    It begin() { return b; }
+    It end() { return e; }
+  };
+
+  // Helper function that creates a reverse iterator for the range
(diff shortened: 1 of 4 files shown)


polytracker/src/taintdag/fdmapping.cpp
```diff
@@ -1,5 +1,28 @@
 #include "taintdag/fdmapping.hpp"
 
+#include <iterator>
+
+namespace {
+
+// Helper structure that allows range-based for iteration
+// on the entries in FDMapping.
+template <typename It> struct RetVal {
```
surovic, line 9: We could maybe use [llvm::make_range](https://llvm.org/doxygen/iterator__range_8h.html) here?

surovic (approved):

polytracker/src/taintdag/fdmapping.cpp
hbrodin: True. It is almost identical implementation, they handle more cases (move etc.). I don't think we have any dependency towards llvm in the runtime part of polytracker. But this is header only and is probably a good idea. Will have a look.

surovic: > I don't think we have any dependency towards llvm in the runtime part of polytracker.

Ah, okay, didn't realize this! I think we should keep it that way then.

hbrodin: We can't use it right away, would still need some code to map it into a reverse iterator. Keep as is?

surovic: Yes.

PR #6434, opened 2022-03-02
Sets tests dependent on tracing to skip.

surovic:


ESultanik (approved):

ESultanik:
You might need to add another skip to this test: https://github.com/trailofbits/polytracker/runs/5392744087?check_suite_focus=true#step:10:26

But I'm fine to merge this now if you want.

surovic:
> You might need to add another skip to this test: https://github.com/trailofbits/polytracker/runs/5392744087?check_suite_focus=true#step:10:26

This test should work. I suspect this comes up because the constants are taken from a release build and CI runs a debug one.

PR #6445, opened 2022-04-19
Fixes poppler docker build

surovic:


examples/Dockerfile-poppler.demo
```diff
@@ -1,117 +1,50 @@
...
-#RUN ${CXX} --instrument-bitcode -f pdftops.bc -o pdftops_track --output-bitcode pdftops_track.bc --libs /build_artifacts/libpoppler.a freetype fontconfig pthread
-
-#RUN get-bc -b pdfunite 
-#RUN ${CXX} --instrument-bitcode -f pdfunite.bc -o pdfunite_track --output-bitcode pdfunite_track.bc --libs /build_artifacts/libpoppler.a freetype fontconfig pthread
-
-#RUN get-bc -b pdfattach 
-#RUN ${CXX} --instrument-bitcode -f pdfattach.bc -o pdfattach_track --output-bitcode pdfattach_track.bc --libs /build_artifacts/libpoppler.a freetype fontconfig pthread
-
-#RUN get-bc -b pdfimages 
-#RUN ${CXX} --instrument-bitcode -f pdfimages.bc -o pdfimages_track --output-bitcode pdfimages_track.bc --libs /build_artifacts/libpoppler.a freetype fontconfig pthread
```
hbrodin, line 113: Any particular reason we don't instrument these? I suggest that we at least include pdftops, that would allow for a comparison with mutool .ps export.

surovic: > Any particular reason we don't instrument these?

I figured we weren't using them and the instrumentation command was the same so I got rid of them.

> I suggest that we at least include pdftops, that would allow for a comparison with mutool .ps export.

Agreed!

PR #6446, opened 2022-04-19
Update README with taint-dag changes

facutuesca:
- Updates README to be in line with taint-dag changes
- Adds `arm64` build instructions
- Bumps version to 4.0.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,6 +299,22 @@ The following tools are required to test and run PolyTracker:
 * [gllvm](https://github.com/SRI-CSL/gllvm) (`go get github.com/SRI-CSL/gllvm/cmd/...`) is used to create whole program
   bitcode archives and to extract bitcode from targets.
 
+### Building on Apple silicon:
+
+Prebuilt Docker images for `polytracker-llvm` are only available for `amd64`. Users with `arm64` systems will have to
+build the image locally and then change `polytracker`'s Dockerfile to point to it:
+```commandline
+$ mkdir repos && cd repos
+$ git clone https://github.com/trailofbits/polytracker
+$ git clone https://github.com/trailofbits/polytracker-llvm
+$ cd polytracker-llvm
+$ DOCKER_BUILDKIT=1 docker build -t trailofbits/polytracker-llvm .
+$ cd ../polytracker
+$ ## Replace the first line of the Dockerfile with "FROM trailofbits/polytracker-llvm:latest" (no quotes)
+$ docker build -t trailofbits/polytracker .
+```
+
+
 ## Building with LLVM Xray Instrumentation
 It's possible to build both Polytracker and target programs with [LLVM's Xray](https://llvm.org/docs/XRay.html)
 instrumentation. Pass the `XRAY_ON` option to Polytracker, and build targets with `--xray-instrument-target` or


README.md
```diff
@@ -70,7 +70,7 @@ polytracker --help
 The `polytracker` script is also a REPL, if run with no command line arguments:
 ```python
 $ polytracker
-PolyTracker (3.0.0)
+PolyTracker (3.1.0)
```
hbrodin, line 73: Suggest that we bump the version to 4.0.0 because this is a non-backwards compatible change with rather large impact on previous behaviour. This should also be reflected in the polytracker.h version.

README.md
```diff
@@ -114,51 +114,39 @@ meant to be tracked. (Note: PolyTracker can in fact track multiple input files
...
-for called_function in entrypoint.calls():
-    # called_function is another function invocation object
-    # associated with a function called from entrypoint
-    print(str(called_function))
-    for region in called_function.taints().regions():
-        # this will iterate over every tainted region operated on by called_function,
-        # including any functions called to from called_function
-        print(f"\t{region.value}")
+trace = PolyTrackerTrace.load("polytracker.tdag")
+# TODO(facutuesca): replace with example using new API
```
hbrodin, line 123: Check with @surovic for examples. He implemented the api-part and probably have some example lying around.

README.md
```diff
@@ -114,51 +114,39 @@ meant to be tracked. (Note: PolyTracker can in fact track multiple input files
...
+# for taint in main_func.taints().regions():
+#     # this iterates over every taint touched by the main function aggregated across all invocations of main
+#     print(f"source={taint.source}, offset={taint.offset}, length={taint.length}, value={taint.value}")
 ```
 
 You can also run an instrumented binary directly from the REPL:
 ```python
 $ polytracker
-PolyTracker (3.0.0)
+PolyTracker (3.1.0)
```
hbrodin, line 143: Same as above

README.md
```diff
@@ -114,51 +114,39 @@ meant to be tracked. (Note: PolyTracker can in fact track multiple input files
...
 https://github.com/trailofbits/polytracker
 Type "help" or "commands"
 >>> trace = run_trace("path_to_binary", "path_to_input_file")
->>> for event in trace:
-...     print(event)
+# TODO(facutuesca): Replace with working example new API
+# >>> for event in trace:
+# ...     print(event)
 ```
```
hbrodin, line 150: Consider only including one example since we dropped the basic block/function tracing and currently only have one use case.

README.md
```diff
@@ -265,7 +253,7 @@ docker build -t trailofbits/polytracker-demo-mupdf -f examples/pdf/Dockerfile-mu
...
 
 `mutool_track` will be build in `/polytracker/the_klondike/mupdf/build/debug`. Running `mutool_track` will output
-`polytracker.db` which contains the information provided by the taint analysis. Its recommended to use this json with
+`polytracker.tdag` which contains the information provided by the taint analysis. Its recommended to use this json with
```
hbrodin, line 256: it is not a json-file, that is some legacy wording

facutuesca: Should I also remove the recommendation to use Polyfile, since it does not support `.tdag`?

README.md
facutuesca: I removed the commented out section, but left the `run_trace` line to have at least one example of the REPL being used

README.md
facutuesca: Bumped to 4.0.0

README.md
facutuesca: Fixed

README.md
hbrodin: Yes, I think so. At least for now, until we've added support for tdag in polyfile.

facutuesca: Fixed

README.md
facutuesca: Added example

README.md
```diff
@@ -295,6 +295,22 @@ The following tools are required to test and run PolyTracker:
...
+### Building on Apple silicon:
+
+Prebuilt Docker images for `polytracker-llvm` are only available for `amd64`. Users with `arm64` systems will have to
+build the image locally and then change `polytracker`'s Dockerfile to point to it:
+```commandline
+$ mkdir repos && cd repos
+$ git clone https://github.com/trailofbits/polytracker
+$ git clone https://github.com/trailofbits/polytracker-llvm
+$ cd polytracker-llvm
+$ DOCKER_BUILDKIT=1 docker build -t trailofbits/polytracker-llvm .
```
hbrodin, line 307: Do you need to explicitly specify `DOCKER_BUILDKIT=1` for it to work?

README.md
```diff
@@ -295,6 +295,22 @@ The following tools are required to test and run PolyTracker:
...
+$ cd ../polytracker
+$ ## Replace the tagged image name on the first line with trailofbits/polytracker:latest
```
hbrodin, line 309: Suggest that we provide the full line e.g. `FROM trailofbits/polytracker:latest`

README.md
facutuesca: Yes, but I'm not sure if this is always the case (I'm running docker on a `multipass` Ubuntu VM, instead of using Docker desktop)

README.md
facutuesca: Fixed! I also realised it should have said `polytracker-llvm` instead of `polytracker`

PR #6447, opened 2022-04-20
Tdag documentation

hbrodin:


docs/tdag.md
```diff
@@ -0,0 +1,105 @@
+# TaintDAG file format
+
+The Taint Directed Acyclic Graph (TaintDAG, TDAG) file format is tailored to facilitate fast recording of taint operations.
+
+It is a binary file format based on sparse files. It consists of a header and a number of subsections. The subsections stores information about:
```
facutuesca, line 5:
```suggestion
It is a binary file format based on sparse files. It consists of a header and a number of subsections. The subsections store information about:
```

docs/tdag.md
```diff
@@ -0,0 +1,105 @@
...
+* taint sources, filename and offset information
+* taint output log, tainted values written to an output file
+* taint graph, the graph of how taint values are unioned from source taint and other unions.
```
facutuesca, line 8:
```suggestion
* taint sources: filename and offset information
* taint output log: tainted values written to an output file
* taint graph: the graph of how taint values are unioned from source taint and other unions.
```

docs/tdag.md
```diff
@@ -0,0 +1,105 @@
...
+
+Because of the sparse layout it is very well suited for mmap'ing directly into the instrumented process address space.
```
surovic, line 10: Maybe "...for mmap'ing directly..." could be replaced with "...for memory mapping (via `mmap()`)...". It's a thing of stylistic preference though.

docs/tdag.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Taint Sources, Unions and Ranges
+Whenever data is read from an input file, the data entering the program is labeled as source taint. Information about which file and at what offset is kept. This is the only way taint can originate in a program.
```
surovic, line 13: "way taint can"  -> "way taints can"

docs/tdag.md
```diff
@@ -0,0 +1,105 @@
...
+
+Whenever an operation is performed where one or more operands are involved a union or range value is created. If the taint labels are adjacent (number wise), e.g. two consecutive source taint bytes, a range is created. If the labels are not adjacent a union of the two labels is created. Unions and ranges occupy the same amount of storage. The main difference is that a range can be extended to become a larger range.
+
+The main motivation for introducing ranges is to allow very efficient membership testing. If a taint label is already included in a range of taint values, the range can be reused. It is possible to walk the tree of unioned labels but it reqiures more computation.
+
+## Affects control flow
+In addition to being Source-, Union- or Range-Taint, each value is also marked if it affects control flow. The basic example is if a value `x` (having label `y`) read from file is compared against another value and a branch is taken based on the result. Whenever the comparison and and branch is executed taint with label `y` is marked as affecting control flow.
```
facutuesca, line 20:
```suggestion
In addition to being Source-, Union- or Range-Taint, each value is also marked if it affects control flow. The basic example is if a value `x` (having label `L`) read from file is compared against another value and a branch is taken based on the result. Whenever the comparison and branch is executed, the taint with label `L` is marked as affecting control flow.
```

docs/tdag.md
```diff
@@ -0,0 +1,105 @@
...
+```C
+struct FDMappingHdr {
+  int32_t fd;
+  uint32_t name_offset;
+  uint32_t name_len;
+  uint32_t prealloc_begin;
+  uint32_t prealloc_end;
+};
+```
+Each of the `FDMappingHdr` structures have an implicit index to which they are refered to in subsequent structures in the tdag.
```
facutuesca, line 54:
```suggestion
Each of the `FDMappingHdr` structures has an implicit index. Subsequent structures in the tdag use that index to refer to each `FDMappingHdr`
```

docs/tdag.md
```diff
@@ -0,0 +1,105 @@
...
+* taint sources, filename and offset information
+* taint output log, tainted values written to an output file
+* taint graph, the graph of how taint values are unioned from source taint and other unions.
+
+Because of the sparse layout it is very well suited for mmap'ing directly into the instrumented process address space.
+
+## Taint Sources, Unions and Ranges
+Whenever data is read from an input file, the data entering the program is labeled as source taint. Information about which file and at what offset is kept. This is the only way taint can originate in a program.
+
+Whenever an operation is performed where one or more operands are involved a union or range value is created. If the taint labels are adjacent (number wise), e.g. two consecutive source taint bytes, a range is created. If the labels are not adjacent a union of the two labels is created. Unions and ranges occupy the same amount of storage. The main difference is that a range can be extended to become a larger range.
```
surovic, line 15: Probably be better if we specify that the operations we have in mind are operations of the instrumented program and the operands need to be tainted for a union or range to be created. Is that right?

surovic: Hmm. Maybe this whole paragraph needs insight into how we do taint analysis. So the previous comment might not be relevant.

docs/tdag.md
```diff
@@ -0,0 +1,105 @@
...
+
+The main motivation for introducing ranges is to allow very efficient membership testing. If a taint label is already included in a range of taint values, the range can be reused. It is possible to walk the tree of unioned labels but it reqiures more computation.
```
surovic, line 17: We probably need to explanation how unions and ranges relate to each other for the "tree of unioned labels" part to make sense.

docs/tdag.md
```diff
@@ -0,0 +1,105 @@
...
+  uint32_t prealloc_end;
+};
+```
+Each of the `FDMappingHdr` structures have an implicit index to which they are refered to in subsequent structures in the tdag.
+```
+[FDMappingHdr][FDMappingHdr]...[FDMappingHdr]
+Index 0        Index 1      ... Index N
+```
+The `fd` indicates the number of the fd as seen in the running program.  
+The `name_offset` and `name_len` specifies offset and length to the filename that was opened and having file descriptor `fd` in the program.
```
facutuesca, line 60:
```suggestion
The `name_offset` and `name_len` specifies offset and length of the filename that was opened and had file descriptor `fd` in the program.
```

docs/tdag.md
surovic: "reqiures" -> "requires"

docs/tdag.md
```diff
@@ -0,0 +1,105 @@
...
+The sinklog is a sequence of records logging what tainted values have been writeen to output files. Each entry in the sinklog index is defined as
+```C
+struct SinkLogEntry {
+  uint8_t fdidx;
+  uint64_t offset;
+  uint32_t label;
+};
+```
+NOTE: The structure is assumed to be packed and occupy `1 + 8 + 4 = 13` bytes.
+In this structure, the `fdidx` member is an index into the `FDMappingHdr` array previously described. The `offset` is current offset in the output file (represented by `fdidx`). Finally the `label` is the taint label associated with the data written and is thus an index into the structure at `tdag_mapping_offset` into the file.
```
facutuesca, line 100:
```suggestion
In this structure, the `fdidx` member is an index into the `FDMappingHdr` array previously described. The `offset` is the offset in the output file (represented by `fdidx`). Finally the `label` is the taint label associated with the data written and is thus an index into the TDAG structure at `tdag_mapping_offset`.
```

docs/tdag.md
```diff
@@ -0,0 +1,105 @@
...
+  63  61        30      0
+```
+Here, `v` and `w` denotes unsigned integers referring to other taint values in the tdag structure. To differentiate between a range and a union the following rule is used:
+```
+v < w => RangeTaint
+w < v => UnionTaint
+w == v => undefined
+```
+### SinkLog
+The sinklog is a sequence of records logging what tainted values have been writeen to output files. Each entry in the sinklog index is defined as
```
facutuesca, line 91:
```suggestion
The sinklog is a sequence of records logging what tainted values have been written to output files. Each entry in the sinklog index is defined as
```

docs/tdag.md
hbrodin: Agree

docs/tdag.md
```diff
@@ -0,0 +1,105 @@
...
+The `name_offset` and `name_len` specifies offset and length to the filename that was opened and having file descriptor `fd` in the program.
+The `prealloc_begin` and `prealloc_end`, if not zero, indicates a source taint range that was preallocated for this file. The idea is to have as many contiguous labels as possible for the same file, aiming at maximising the number of ranges generated.
+
+### SourceTaint, UnionTaint and RangeTaint - the actual TDAG
+At `tdag_mapping_offset` there are `tdag_mapping_count` uint64_t entries. Each entry denotes either SourceTaint, UnionTaint or RangeTaint. Their relative index is the taint label. Index zero is unused as it denotes 'not tainted'. The general layout of these 64-bit values are:
+```
+| x y zzz...z |
+  63        0
+``` 
+Bits `x` and `y` are common for the three kinds of taint values. The value `y` is set to one if the taint represents control flow and zero if not. The value `x` is set to one to indicate that it is a source taint and
```
facutuesca, line 69:
```suggestion
Bits `x` and `y` are common for the three kinds of taint values. The value `y` is set to one if the taint affects control flow and zero if not. The value `x` is set to one to indicate that it is a source taint and
```

facutuesca (commented):
This is great! The explanation and the details are super clear.

I suggested some changes to clarify a few points, but feel free to ignore any of them if you feel they don't improve clarity

docs/tdag.md
surovic, line 20: I'd suggest that "...if a value `x` (having label `y`)..." -> "if a value with taint label `y`"  is probably better since we don't refer to `x` anywhere. Same thing below with `v` and `w`.  "...read from a file is compared..." -> "is read from a file and compared...". "and and" -> "and" ; "taint with label" -> "taint label" ; "executed" -> "executed, ".

docs/tdag.md
```diff
@@ -0,0 +1,105 @@
...
+  uint64_t tdag_mapping_offset;
+  uint64_t tdag_mapping_size;
+  uint64_t sink_mapping_offset;
+  uint64_t sink_mapping_size;
+};
+```
+Each offset is relative to file (or memory mapping) start.
+
+### FDMappingHdr
+At `fd_mapping_offset` there are `fd_mapping_count` `FDMappingHdr` structures.
```
surovic, line 44: "At `fd_mapping_offset` there is an array of `FDMappingHdr` structures. Length of the array is given by `fd_mapping_count`. " maybe?

surovic: We should also mention that "FD" means file descriptor.

docs/tdag.md
surovic: "having" -> "has"

docs/tdag.md
```diff
@@ -0,0 +1,105 @@
...
+
+The main motivation for introducing ranges is to allow very efficient membership testing. If a taint label is already included in a range of taint values, the range can be reused. It is possible to walk the tree of unioned labels but it reqiures more computation.
+
+## Affects control flow
+In addition to being Source-, Union- or Range-Taint, each value is also marked if it affects control flow. The basic example is if a value `x` (having label `y`) read from file is compared against another value and a branch is taken based on the result. Whenever the comparison and and branch is executed taint with label `y` is marked as affecting control flow.
+
+Affects control flow propagates through unions and ranges meaning that if a value `v`, having label `w` where `w` is a union or range, affects control flow. Then each taint label represented by `w` is in turn marked as affecting control flow.
```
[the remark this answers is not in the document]
hbrodin: I agree it all becomes a bit abstract when there are no pictures. I've made slight modifications to maybe clarify a bit. Unfortunately the suggestion you made is not correct. `z` does not necessarily affect control flow, even though `v` does. But if `z` does affect control flow, then both `w` and `v` affects control flow.

docs/tdag.md
hbrodin: It is mentioned just below, is that good enough?

docs/tdag.md
```diff
@@ -0,0 +1,105 @@
...
+  uint32_t prealloc_begin;
+  uint32_t prealloc_end;
+};
+```
+Each of the `FDMappingHdr` structures have an implicit index to which they are refered to in subsequent structures in the tdag.
+```
+[FDMappingHdr][FDMappingHdr]...[FDMappingHdr]
+Index 0        Index 1      ... Index N
+```
+The `fd` indicates the number of the fd as seen in the running program.
```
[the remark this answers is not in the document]
hbrodin: It is directly the file descriptor as returned by `open`, (`fopen` returns a FILE*, which of course can be converted to an int...).

PR #6449, opened 2022-04-25
Add support for `putc`/`fputc`

facutuesca:
This PR adds wrappers for `putc()` and `fputc()`. It also adds an overload for `PolyTracker::taint_sink()`, where instead of taking a data buffer (getting each byte's taint label), it takes a single label and uses that for the whole range.


polytracker/custom_abi/dfsan_abilist.txt
```diff
@@ -3054,6 +3054,7 @@ fun:ptrace=uninstrumented
 fun:ptsname=uninstrumented
 fun:ptsname_r=uninstrumented
 fun:putc=uninstrumented
+fun:putc=custom
```
hbrodin, line 3057: Isn't it enough to only have a single `fun:putc=custom` entry? I realise these files contain a mix of such statements, but if it works I think we should make a slight improvement by only having one entry.

polytracker/custom_abi/polytracker_abilist.txt
```diff
@@ -3054,6 +3054,7 @@ fun:ptrace=uninstrumented
 fun:ptsname=uninstrumented
 fun:ptsname_r=uninstrumented
 fun:putc=uninstrumented
+fun:putc=custom
```
hbrodin, line 3057: Same as the comment for `dfsan_abilist.txt`

hbrodin (approved):
Great job! LGTM.

polytracker/custom_abi/dfsan_abilist.txt
facutuesca: It also needs the `uninstrumented` entry. Reading [this](https://github.com/trailofbits/polytracker-llvm/blob/123f8d2b2bdfcc0d2577e8836f2e893e823037c5/llvm/lib/Transforms/Instrumentation/DataFlowSanitizer.cpp#L169-L175) and [this](https://clang.llvm.org/docs/DataFlowSanitizer.html#abi-list), it seems that by default DFSAN treats every function as instrumented (which doesn't work for library functions like `fputc`), and if you want to avoid that you need to specify both `uninstrumented` plus the annotation that specifies how to handle it (`discard`, `functional` or `custom`).

Just in case, I tried compiling without `uninstrumented` and it did fail:
```
Instrumenting:  91% |██████████████████  | 11/11 [00:00, inf functions/s]       
/usr/bin/ld: a_instrumented.o: in function `foo(int)':
/polytracker/build/fputc.cpp:9: undefined reference to `dfs$fputc'
clang-13: error: linker command failed with exit code 1 (use -v to see invocation)
```

polytracker/custom_abi/polytracker_abilist.txt
facutuesca: same as above

polytracker/custom_abi/dfsan_abilist.txt
hbrodin: 👍

PR #6450, opened 2022-04-27
Split pass refactor

facutuesca:
This is a small refactor that moves code used for debugging (logging the function names during the BB splitting pass) into its own separate function.
It also puts the code used to get the function name inside the `#ifdef DEBUG_INFO` guard, since the function name won't be used for anything else.

The logic for logging was changed (see `log_function()`), since the previous version left empty strings when no `BasicBlock` name was found:
### Before
```
root@c6c5529074d5:/tmp/tmp.9wFNLVmLp4/build/bin# ./polybuild_script++ --instrument-target -g -o a.out a.cpp 
Bitcode file extracted to: a.bc.
Instrumenting:   0% |                    | 1/10 [00:00, inf functions/s] after call to fopen
 after call to fseek
 after call to fgetc
 after call to fopen
 after call to fclose
 after call to fclose
Instrumenting:  50% |██████████          | 6/10 [00:00, inf functions/s] after call to _Z3fooi
Instrumenting:  90% |██████████████████  | 10/10 [00:00, inf functions/s] 
```
### After
```
root@c6c5529074d5:/tmp/tmp.9wFNLVmLp4/build/bin# ./polybuild_script++ --instrument-target -g -o a.out a.cpp 
Bitcode file extracted to: a.bc.
Instrumenting:   0% |                    | 1/10 [00:00, inf functions/s] Unnamed block after call to fopen
 Unnamed block after call to fseek
 Unnamed block after call to fgetc
 Unnamed block after call to fopen
 Unnamed block after call to fclose
 Unnamed block after call to fclose
Instrumenting:  50% |██████████          | 6/10 [00:00, inf functions/s] Unnamed block after call to _Z3fooi
Instrumenting:  90% |██████████████████  | 10/10 [00:00, inf functions/s]  
```

The best way to review is commit-by-commit, since it shows more clearly what is being changed.


polytracker/src/passes/bb_splitting_pass.cpp
```diff
@@ -41,34 +81,10 @@ BBSplittingPass::analyzeBasicBlock(BasicBlock &basicBlock) const {
...
-      if (fname.length() == 0 || bb->hasName() || includeFunctionName) {
-        if (bb->hasName()) {
-          llvm::errs() << " " << bb->getName().data();
-        }
-        if (includeFunctionName) {
-          llvm::errs() << " after call to " << fname;
-        }
-        llvm::errs() << "\n";
-      }
+      log_function(bb, get_function_name(call));
```
surovic, line 87: Do you think you could move the conditionally compiled code out if `analyzeBasicBlock` and provide an empty `log_function` for release builds and this one for debug?

surovic (changes requested):

polytracker/src/passes/bb_splitting_pass.cpp
facutuesca: What are the advantages of having it like that? I'm partial to this approach since having:
```c++
void log_function() { // implementation }

std::vector analyzeBasicBlock() {
// ... some code

#ifdef DEBUG_INFO
    log_function();
#endif

///... other code
}
```
makes it clear to readers of `analyzeBasicBlock` that `log_function` only matters for debug builds. 
If there were several places where `log_function` is called, then I would definitely agree to move the debug check there (to avoid duplication and noise). But since this is a helper function that is only used once, I think this way is a bit clearer.

What do you think?
```

surovic: The advantage would be that we could later re-use it in other places in polytracker. But at that point it would probably be better just to use a proper logging library like [spdlog](https://github.com/gabime/spdlog).

surovic (approved):

PR #6455, opened 2022-04-29
Small refactor for `PolytrackerPass`

facutuesca:
This replaces ~50 lines of code that implement a custom progress bar with a small header-only 3rd party library that does the same

https://user-images.githubusercontent.com/5762120/166220095-f2a2be9d-b380-4053-a628-9c29d6583c2c.mov

### DIFF
--- a/.github/workflows/pull_req.yml
+++ b/.github/workflows/pull_req.yml
@@ -11,6 +11,8 @@ jobs:
         example: [mupdf,poppler,qpdf]
     steps:
       - uses: actions/checkout@v2
+        with:
+          submodules: true
       - name: Build the base image
         run: docker build . --file Dockerfile --tag trailofbits/polytracker --no-cache
       - name: Build the example Docker image
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "third_party/run-clang-format"]
 	path = third_party/run-clang-format
 	url = https://github.com/Sarcasm/run-clang-format.git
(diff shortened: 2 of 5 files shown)


surovic:
I honestly question the need for this code. If it is indeed desirable, isn't there a library that does exactly this?

facutuesca:
> I honestly question the need for this code. If it is indeed desirable, isn't there a library that does exactly this?
> 
> Edit: perhaps [FTXUI](https://github.com/ArthurSonzogni/FTXUI)

I would guess it's not strictly needed, just a nice to have. Do you think I should remove it on this PR?

surovic:
> I would guess it's not strictly needed, just a nice to have. Do you think I should remove it on this PR?

I would be happier if it wasn't there. But I'd wait for @ESultanik and @hbrodin opinons.

facutuesca:
> I honestly question the need for this code. If it is indeed desirable, isn't there a library that does exactly this?
> 
> Edit: perhaps [FTXUI](https://github.com/ArthurSonzogni/FTXUI)

This is what it could look like using this [small header-only library](https://github.com/p-ranav/indicators):

https://user-images.githubusercontent.com/5762120/166204382-6a37a373-393b-4da8-84ec-9241595e2903.mov

surovic:
This already looks like a much better solution imo.

facutuesca:
> This already looks like a much better solution imo.

@surovic I replaced the progress bar implementation with the new library, let me know how it looks

surovic (approved):

PR #6458, opened 2022-05-03
Removes nlohmann-json dependency

### LINKED ISSUE #22: Prominently include license for nlohmann-json
We currently use (at least) `CRoaring` and `nlohmann-json` as third-party dependencies. These are released under the Apache v2 license and an MIT license, respectively.

We should make sure that we are distributing the third-party packages licenses, as required by them. Perhaps add a top-level `LICENSES.thirdparty` directory with license files there.

surovic:
Resolves #22


hbrodin (approved):
LGTM

PR #6459, opened 2022-05-04
Fix a type issue that causes infinite loop on aarch64

hbrodin:
This is what remains from #6388, wanted to remove that PR. Once this is merged we can do so.


surovic (approved):
What's the difference on aarch64?

hbrodin:
> 

Don't remember exactly. The issue is that EOF is of type int. On aarch64 the comparison was always false. So, maybe some endianess issue...

PR #6465, opened 2022-05-12
Move `dumptdag.py` functionality into core API

surovic:
The PR moves functionality from [dumptdag.py](https://github.com/trailofbits/polytracker/blob/master/polytracker/dumptdag.py) into the core API under an updated [cavities](https://github.com/trailofbits/polytracker/blob/03c05d057defa38c3bb3fff0c18a974307b49e59/polytracker/mapping.py#L157) command and new [info](https://github.com/trailofbits/polytracker/blob/03c05d057defa38c3bb3fff0c18a974307b49e59/polytracker/taint_dag.py#L445) command. As a consequence `dumptdag.py` is removed and replaced with the new API commands ([file_cavity_detection.py](https://github.com/trailofbits/polytracker/blob/a4e387d63da3ee71985c61f105eec8ef125ecddb/build_in_docker/file_cavity_detection.py#L356)).


tests/test_tdag.py
```diff
@@ -1,42 +1,6 @@
...
-from polytracker import dumptdag, taint_dag, ProgramTrace
-from os import getcwd
+from polytracker import taint_dag, ProgramTrace
 from typing import cast
-from .data import TEST_DATA_DIR, TEST_DATA_PATH, TEST_RESULTS_DIR
-from polytracker.containerization import CAN_RUN_NATIVELY
-
-
-@pytest.mark.program_trace("test_tdag.cpp")
-def test_dumptdag(program_trace: ProgramTrace):
```
hbrodin, line 10: Do we have any other test that checks these details? If not, I think we should rewrite this test case to use the new api.

polytracker/mapping.py
```diff
@@ -3,156 +3,191 @@
...
+                        if isinstance(rn, TDSourceNode):
+                            yield rl
+                        else:
+                            stack.append(rl)
+
+        result: Dict[Path, List[CavityType]] = defaultdict(list)
+
+        for p, h in self.tdfile.fd_headers:
+            begin = h.prealloc_label_begin
+            end = h.prealloc_label_end
```
hbrodin, line 93: We previously had the discussion about not always preallocating ranges. Would it make sense to add an assert here of some sort? E.g. assert that `end != 0 and begin !=0`. I guess it might not work because output files are in this enumeration as well and would typically have initial size zero.

polytracker/mapping.py
```diff
@@ -3,156 +3,191 @@
...
+        if start is not None:
+            ranges.append((start, len(m) - 1))
+        return ranges
+
+    def file_cavities(self) -> Dict[Path, List[CavityType]]:
+        seen: Set[int] = set()
+
+        def source_labels_not_affecting_cf(label: int) -> Generator[int, None, None]:
+            stack = [label]
+            while len(stack) > 0:
```
hbrodin, line 46:
```suggestion
            while stack:
```

polytracker/mapping.py
```diff
@@ -3,156 +3,191 @@
...
+                    for rl in range(n.first, n.last + 1):
+                        # NOTE: One could skip decoding here, but then we could end up with really long ranges
+                        # being added the labels that really does nothing except cause overhead...
+                        rn = self.tdfile.decode_node(rl)
+                        if rn.affects_control_flow:
+                            continue
+                        if isinstance(rn, TDSourceNode):
+                            yield rl
+                        else:
+                            stack.append(rl)
```
hbrodin, line 87: Maybe we could add the `(rl, rn)` here and have a check on line 54 that checks `isinstance(lbl, int)` before doing `decode_node`? That would prevent decoding a taint node twice. Might be worth experimenting if it increases performance.

hbrodin: Also, if this is done, one should probably do the same for `n.left` and `n.right` when it is a TDUnionNode.

build_in_docker/file_cavity_detection.py
```diff
@@ -351,12 +351,7 @@ def file_cavity_detection(
             print(f"Error when processing {filename}")
             return f"{filename},-3,-3\n"
 
-        command = [
-            "python3",
-            str(SCRIPTDIR / "../polytracker/dumptdag.py"),
-            str(dst_tdag),
-            str(tool.container_input_path(filename)),
-        ]
+        command = ["polytracker", "cavities", str(dst_tdag)]
```
hbrodin, line 354: This is a slight change in behavior, right? Previously we only computed cavities for the processed file. If I don't misread, now we would get cavities for all files being processed by the command.

polytracker/mapping.py
```diff
@@ -3,156 +3,191 @@
...
+        raise NotImplementedError()
+
+    def marker_to_ranges(self, m: bytearray) -> List[CavityType]:
+        ranges = []
+        start = None
+        for i, v in enumerate(m):
+            if v == 0:
+                if start is None:
+                    start = i
+            if v == 1:
```
hbrodin, line 33:
```suggestion
            else:
```

polytracker/mapping.py
```diff
@@ -3,156 +3,191 @@
...
 
     def __init_arguments__(self, parser):
-        parser.add_argument("POLYTRACKER_DB", type=str, help="the trace database")
+        parser.add_argument("POLYTRACKER_TF", type=str, help="the trace file")
 
     def run(self, args):
-        mapping = InputOutputMapping(PolyTrackerTrace.load(args.POLYTRACKER_DB)).mapping
-
-        print(mapping)
+        raise NotImplementedError()
```
hbrodin, line 130: Any reason we couldn't keep this? I mean, if you wanted to get a mapping even though it is slower. Or does it just not make sense anymore?

hbrodin (commented):
Great job @surovic! Most of the comments are nitpicks or suggestions. The one thing I wanted to make sure is that the potential change in behaviour (cavities for all input files) doesn't cause any issues.

polytracker/mapping.py
surovic: I plan on re-implementing it as my next step. Take out the DFS walk out of cavity detection and make it it's own method and use it in both `cavities` and `mapping`.

build_in_docker/file_cavity_detection.py
surovic: All file descriptors tracked in the `.tdag` file, but yes. I think it shouldn't be a problem, unless the tool introduces taint sources from a file not specified by `filename`. That doesn't happen with the tools we currently support in the `file_cavity_detection.py` script, right? Correct me if I'm wrong though! 

One thing I could do is save us a shell spawn, detect cavities using API and filter out results for `filename`.

polytracker/mapping.py
surovic: I'm not sure I understand what you mean here. Can you write a suggestion with the changes?

polytracker/mapping.py
surovic: There's a test for length on line 96 :eyes: Is it not enough?

tests/test_tdag.py
surovic: There's test for the API version in the same file IIRC.

polytracker/mapping.py
hbrodin: My point is that if we change the implementation to not require preallocated ranges this would stop working. Maybe that's what we have tests for...

polytracker/mapping.py
hbrodin, line 87:
```suggestion
                            stack.append((rl, rn))
```

polytracker/mapping.py
```diff
@@ -3,156 +3,191 @@
...
+            stack = [label]
+            while len(stack) > 0:
+                lbl = stack.pop()
+
+                if lbl in seen:
+                    continue
+
+                seen.add(lbl)
+
+                n = self.tdfile.decode_node(lbl)
```
hbrodin, line 54:
```suggestion
                # Conceptually like this, not sure valid python exactly.
                if isinstance(n, int):
                    n = self.tdfile.decode_node(lbl)
                else:
                    (lbl, n) = lbl
                
```

polytracker/mapping.py
```diff
@@ -3,156 +3,191 @@
...
+                if isinstance(n, TDSourceNode):
+                    yield lbl
+
+                elif isinstance(n, TDUnionNode):
+                    nl = self.tdfile.decode_node(n.left)
+                    if not nl.affects_control_flow:
+                        if isinstance(nl, TDSourceNode):
+                            yield n.left
+                        else:
+                            stack.append(n.left)
```
hbrodin, line 68: Same concept as line 87

polytracker/mapping.py
```diff
@@ -3,156 +3,191 @@
...
+
+                    nr = self.tdfile.decode_node(n.right)
+                    if not nr.affects_control_flow:
+                        if isinstance(nr, TDSourceNode):
+                            yield n.right
+                        else:
+                            stack.append(n.right)
```
hbrodin, line 75: Same concept as line 87

polytracker/mapping.py
hbrodin: I've made a few suggestions. Not sure it is valid python code, but conceptually that's what I mean. Makes sense?

build_in_docker/file_cavity_detection.py
hbrodin: I think your last point would be a great addition!

hbrodin (approved):

build_in_docker/file_cavity_detection.py
surovic: Turns out the implementing timeouts for threads is more work than it's worth. I take  my suggestion back :sweat_smile:

polytracker/taint_dag.py
```diff
@@ -120,23 +121,23 @@ def __init__(self, buffer: mmap) -> None:
         self.source_index_bits = 8
         self.source_offset_mask = (1 << 54) - 1
 
-        self.buffer = buffer
+        self.buffer = mmap(file.fileno(), 0, prot=PROT_READ)
```
hbrodin, line 124: Just out of curiosity, why the move of mmap into the init-function?

polytracker/mapping.py
```diff
@@ -4,24 +4,59 @@
...
+from typing import Dict, List, Set, Tuple, Iterator
 from tqdm import tqdm
-from mmap import mmap, PROT_READ
 
 from .plugins import Command
-from .taint_dag import TDFile, TDRangeNode, TDSourceNode, TDUnionNode
+from .taint_dag import TDFile, TDNode, TDRangeNode, TDSourceNode, TDUnionNode
 
 
+LabelType = int
```
hbrodin, line 14: Good move with the types!
I suggest that we take it one step further using https://docs.python.org/3/library/typing.html#newtype

polytracker/mapping.py
```diff
@@ -4,24 +4,59 @@
...
 CavityType = Tuple[OffsetType, OffsetType]
 
 
 class InputOutputMapping:
     def __init__(self, f: TDFile):
         self.tdfile: TDFile = f
 
-    def mapping(self) -> Dict[OffsetType, Set[OffsetType]]:
-        raise NotImplementedError()
+    def dfs_walk(
```
hbrodin, line 24: Technically, this is a BFS, right?

polytracker/mapping.py
```diff
@@ -4,24 +4,59 @@
...
+
+            elif isinstance(n, TDUnionNode):
+                stack.append(n.left)
+                stack.append(n.right)
+
+            elif isinstance(n, TDRangeNode):
+                stack.extend(range(n.first, n.last + 1))
+
+    def mapping(self) -> Dict[FileOffsetType, Set[FileOffsetType]]:
+        result: Dict[FileOffsetType, Set[FileOffsetType]] = defaultdict(set)
```
hbrodin, line 51:
```suggestion
        seen = set()
        result: Dict[FileOffsetType, Set[FileOffsetType]] = defaultdict(set)
```

polytracker/mapping.py
```diff
@@ -4,24 +4,59 @@
...
+        for s in tqdm(list(self.tdfile.sinks)):
+            for _, n in self.dfs_walk(s.label):
```
hbrodin, line 53:
```suggestion
            for _, n in self.dfs_walk(s.label, seen):
```

polytracker/mapping.py
hbrodin: I think the `seen` can be reused across the full iteration, that could potentially yield some good speedup. If we've seen a node once we have done all we can with it.

polytracker/mapping.py
```diff
@@ -83,17 +91,24 @@ def source_labels_not_affecting_cf(label: int) -> Generator[int, None, None]:
...
                 else:
-                    for source in source_labels_not_affecting_cf(sink.label):
-                        marker[source - begin] = 1
+                    for lbl, n in self.dfs_walk(s.label, seen):
+                        if isinstance(n, TDSourceNode):
+                            marker[lbl - begin] = 1
+                        elif n.affects_control_flow:
+                            if isinstance(n, TDUnionNode):
+                                seen.add(n.left)
+                                seen.add(n.right)
```
hbrodin, line 109: Clever!

polytracker/taint_dag.py
```diff
@@ -195,7 +196,7 @@ def read_sink(self, offset: int) -> TDSink:
         return result
 
     @property
-    def sinks(self) -> Generator[TDSink, None, None]:
+    def sinks(self) -> Iterator[TDSink]:
         assert self.header.sink_mapping_offset > 0
```
hbrodin, line 200: Suggest that we move the assertions to the __init__-function to get a early error if there is something wrong with the tdag-file.

polytracker/taint_dag.py
```diff
@@ -174,7 +175,7 @@ def decode_node(self, label: int) -> TDNode:
                 return TDRangeNode(v1, v2, affects_cf)
 
     @property
-    def nodes(self) -> Generator[TDNode, None, None]:
+    def nodes(self) -> Iterator[TDNode]:
         assert self.header.tdag_mapping_offset > 0
         assert self.header.tdag_mapping_size > 0
```
hbrodin, line 180: Suggest that we move the assertions to the __init__-function to get a early error if there is something wrong with the tdag-file.

polytracker/mapping.py
```diff
@@ -4,24 +4,59 @@
...
 
 
 class InputOutputMapping:
     def __init__(self, f: TDFile):
         self.tdfile: TDFile = f
 
-    def mapping(self) -> Dict[OffsetType, Set[OffsetType]]:
-        raise NotImplementedError()
+    def dfs_walk(
+        self, label: LabelType, seen: Set[LabelType] = set()
```
hbrodin, line 25: Isn't the `seen = set()` part affected by https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments ?

hbrodin (commented):
The one thing I would check is the default argument `seen` in `dfs_walk`. There are some small comments, but nothing serious. Great job!

polytracker/taint_dag.py
surovic: It's because I had a bunch of places where I've copy-pasted `open()` -> `mmap()` -> `TDFile()` code. So I moved at least the `mmap()` into the `__init__`.

polytracker/mapping.py
surovic: This causes a few headaches, like excessive typecasting. For example with `range(n.first, n.last + 1)` when handling `TDRangeNode`. Mypy doesn't recognize `range` as the built-in function and thinks it's a type. Moreover `NewType` is said to introduce runtime overhead. I think we're fine with just the type aliases so far.

polytracker/mapping.py
surovic: Thanks! TIL!

polytracker/mapping.py
hbrodin: 👍

PR #6478, opened 2022-08-25
Adds `stdout` and `stderr` as taint sinks

### LINKED ISSUE #6477: Add `stdout`, `stderr` as optional taint sinks

surovic:
Resolves #6477


polytracker/include/taintdag/taint.hpp
```diff
@@ -46,7 +46,7 @@ const source_index_t source_index_mask = max_source_index;
 
 // Use the remaining bits for source file offset
 const size_t source_offset_bits = storage_bits - source_index_bits - 2;
-using source_offset_t = storage_t;
+using source_offset_t = int64_t;
```
hbrodin, line 49: Why change? We don't touch stdin at this point, right? I don't think it is a problem, just curious. If it is the output position we wan't to change there is a definition for that here: https://github.com/trailofbits/polytracker/blob/6477-add-stdout-stderr-as-optional-taint-sinks-behind-a-cli-flag/polytracker/include/taintdag/taint_sink_log.hpp#L9

polytracker/src/polytracker/main.cpp
```diff
@@ -96,6 +98,9 @@ void polytracker_start(func_mapping const *globals, uint64_t globals_count,
            "instrumentation.\n");
   }
 
+  get_polytracker_tdag().open_file(fileno(stdout), "/dev/stdout");
+  get_polytracker_tdag().open_file(fileno(stderr), "/dev/stderr");
+
```
hbrodin, line 103: I suggest we guard these statements and check that the return of fileno is >= 0 (in case they are closed for some reason).

hbrodin (commented):
Overall, looks good! Suggest we add two if-statements in case stdout or stderr is closed.
Consider making it a config option what to log. OTOH we don't limit any other logging so... maybe just keep as is and do all of that consistently instead. Up to you!

polytracker/include/taintdag/taint.hpp
surovic: The reason I changed it for both sources and sinks is that the functions that, introduce taints (`pread()`) and sink taints (`fwrite()`) both use signed integers to represent offset. Don't know why sources do, but they do :shrug:

hbrodin: OK, makes sense. Just hoping I haven't made any assumptions about having unsigned types...

surovic:
I've added env var guards for the sinks and a test.

tests/test_stdio.py
```diff
@@ -0,0 +1,26 @@
...
+    monkeypatch.setenv("POLYTRACKER_STDERR_SINK", "1")
+
+
+@pytest.mark.program_trace("test_stdio.cpp")
+def test_tdfile(set_env_vars, program_trace: ProgramTrace):
+    assert isinstance(program_trace, taint_dag.TDProgramTrace)
+    stdout = Path("/dev/stdout")
+    stderr = Path("/dev/stderr")
+    headers = list(program_trace.tdfile.fd_headers)
+    outputs = list(map(lambda h: h[0], headers))
```
hbrodin, line 19: I was confused for a moment here. There should only be two outputs, correct? After more careful reading, I noticed that outputs contain all fd-headers, not only outputs. Consider changing the name to  something like 'processed_filenames' (or something better you come to think of).

hbrodin (approved):
LGTM! Just a minor comment about naming in the test case.

PR #6481, opened 2022-08-29
polytracker: fixes nullptr deref when printing stream sink env vars

surovic:


polytracker/src/polytracker/main.cpp
```diff
@@ -58,8 +58,13 @@ void polytracker_parse_env() {
     get_polytracker_db_name() = pdb;
   }
 
-  get_polytracker_stdout_sink() = getenv("POLYTRACKER_STDOUT_SINK");
-  get_polytracker_stderr_sink() = getenv("POLYTRACKER_STDERR_SINK");
+  if (auto out = getenv("POLYTRACKER_STDOUT_SINK")) {
+    get_polytracker_stdout_sink() = out;
```
hbrodin, line 62: Before touching the `polytracker_stderr_sink` you need to construct it. Sorry, the macros are a bit weird. You should probably add a 
```c++
DO_EARLY_DEFAULT_CONSTRUCT(std::string, polytracker_stdout_sink)
```
line to `polytracker_start` for `polytracker_stdout_sink` and `polytracker_stderr_sink`, like what is done for `polytracker_db_name` on line 112. `DECLARE_EARLY_CONSTRUCT` doesn't actually run the ctor... 😭

hbrodin (approved):
LGTM. I haven't run the code, but I think this is the way to go.

PR #6495, opened 2022-10-07
Update Python version to support new version of Ubuntu in polytracker-llvm

josephsweeney:
Pairs with https://github.com/trailofbits/polytracker-llvm/pull/13

Needed because Ubuntu-jammy doesn't have python3.8 but has 3.10 instead.


kaoudis:
This is an intermediate step so we can get to Ubuntu jammy and ensures that Polytracker works with Python3.10. Once we are on jammy, we should remove the deadsnakes PPA and just pull 3.10 from the regular jammy repos.

kaoudis:
tests are currently failing locally in act on this branch in 3.10 I think due to distutils, c.f. https://peps.python.org/pep-0632/

.github/workflows/dockerimage.yml
```diff
@@ -7,13 +7,13 @@ jobs:
...
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: ['3.7', '3.8', '3.9', '3.10']
     steps:
       - uses: actions/checkout@v2
         with:
           submodules: true
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v2
```
kaoudis, line 16: This allows us to use Pythons > 3.9 and allows `act` to work locally without changing the base container img `act` builds in.

ESultanik (approved):

PR #6498, opened 2022-11-01
Minor CMake refactor

surovic:
This PR gets rid of most CMake code that sets global variables like `CMAKE_CXX_FLAGS`.


polytracker/src/CMakeLists.txt
```diff
@@ -50,7 +35,21 @@ set(TAINTDAG_SOURCES
     ${TAINTDAG_DIR}/fnmapping.cpp
     ${TAINTDAG_DIR}/fntrace.cpp)
 
-add_library(Polytracker STATIC ${POLY_SOURCES} ${TAINT_SOURCES}
+add_library(Polytracker STATIC ${POLYTRACKER_SOURCES} ${TAINT_SOURCES}
                                ${TAINTDAG_SOURCES})
+
+set(POLYTRACKER_COMPILE_OPTS -fno-exceptions -fno-rtti -Wall -Werror
+                             -stdlib=libc++ -fPIC)
```
hbrodin, line 42: Just making sure, the `set(CMAKE_POSITION_INDEPENDENT_CODE ON)` is replaced by the `-fPIC`, right?

polytracker/src/CMakeLists.txt
```diff
@@ -50,7 +35,21 @@ set(TAINTDAG_SOURCES
...
+
+if(XRAY_ON)
+  list(APPEND POLYTRACKER_COMPILE_OPTS -fxray-instrument
+       -fxray-instruction-threshold=1)
+endif()
+
+target_include_directories(Polytracker PRIVATE ${POLYTRACKER_CXX_INCLUDE})
+target_compile_options(Polytracker PRIVATE ${POLYTRACKER_COMPILE_OPTS})
+target_link_options(Polytracker PRIVATE -stdlib=libc++)
```
hbrodin, line 51: Should it be `PUBLIC`? We want to use the same stdlib across the entire final binary, right? Maybe that will happen anyway somehow?
If we made it public here (and the `target_compile_options` as well), I believe we wouldn't have to add e.g. https://github.com/trailofbits/polytracker/pull/6498/files#diff-ca270b0ff3ff8ec2df04556b61ade25034e5d38c1c0928f14db37bb63b8d4d8fR18-R21

hbrodin (approved):
Great job @surovic! I'm glad you are doing the hard work on cleaning up our build!
I added a couple of comments that you might want to consider. Your CMAKE skills are probably more up to date than mine so I might be off in my comments.

polytracker/src/CMakeLists.txt
surovic: Yes.

polytracker/src/CMakeLists.txt
surovic: You're probably right, but part of the point of this PR is to make flag settings as local as possible, because our requirements for code is different in different places e.g. `passes` vs `taintdag`. IIRC there's a better way to do the build that @pgoodman suggested before, which I want to do once I get to moving DFSAN from `polytracker-llvm` to `polytracker`. This PR is meant to just fix the most pressing matters.

PR #6499, opened 2022-11-03
Fix nitro nitf example

surovic:


kaoudis (approved):
Seems like these changes are either a) merging in commits from the main branch or b) fixing things up to respect the --target option I missed. Thank you so much for doing this. Merging this in so I can clean up some more of the stuff I didn't need, on top of it :D

PR #6500, opened 2022-11-04
Henrik/section

hbrodin:
Updated section PR. Rewrite of the tdag output file handling to a more structured section-based approach.
Still a work in progress.  At this point, all tests pass.


mypy.ini
```diff
@@ -1,5 +1,12 @@
 [mypy]
 ignore_missing_imports = True
+exclude = (?x) (
```
surovic, line 3: Seems like a workaround to something. Why was this added?

polytracker/include/taintdag/util.h
```diff
@@ -0,0 +1,46 @@
...
+
+inline void dump_range(std::string name, std::span<uint8_t> range) {
+  auto begin = reinterpret_cast<uintptr_t>(&*range.begin());
+  auto end = reinterpret_cast<uintptr_t>(&*range.end());
+  printf("Name: %s begin: %lx end: %lx\n", name.data(), begin, end);
+}
+} // namespace util
+
+// TODO(hbrodin): Quick and dirty impl of map. Could be considerably improved.
+template <typename T, typename F>
```
surovic, line 39: This is just my opinion. But I think uses of this function singnificantly reduce readability in exchange for a few tens of lines of code that most people will immediately recognize and skim over. If we _really_ want to minimize our LoC, let's get a library for a proper monadic interface to `std::variant`

polytracker/mapping.py
```diff
@@ -73,49 +73,71 @@ def marker_to_ranges(self, m: bytearray) -> List[CavityType]:
                     ranges.append((start, i))
                     start = None
         if start is not None:
-            ranges.append((start, len(m) - 1))
+            ranges.append((start, len(m)))
         return ranges
 
     def file_cavities(self) -> Dict[Path, List[CavityType]]:
```
surovic, line 79: Was this tested for equivalent performance? We had issues with improperly implemented DFS traversals before.

polytracker/src/taintdag/polytracker.cpp
```diff
@@ -7,212 +7,165 @@
...
 #include <sanitizer/dfsan_interface.h>
 
 #include <sys/stat.h>
 
-#include "taintdag/error.hpp"
+#include "taintdag/error.h"
 #include "taintdag/fnmapping.h"
 #include "taintdag/fntrace.h"
 
 namespace fs = std::filesystem;
```
surovic, line 19: We should probably remove or make this more local.

tests/conftest.py
```diff
@@ -13,6 +13,11 @@ def pytest_configure(config):
         "program_trace: mark the C/C++ source file to be automatically compiled, instrumented, and run for the test",
     )
 
+    config.addinivalue_line(
```
surovic, line 16: I think this shouldn't be needed anymore.

unittests/src/taintdag/bitmap_section.cpp
```diff
@@ -0,0 +1,201 @@
+#include <random>
+
+#include <catch2/catch.hpp>
+
+#include "taintdag/bitmap_section.h"
+
+#include "error_exit_helper.h"
```
surovic, line 7: Should be replaced with `utils.h`

unittests/src/taintdag/error_exit_helper.h
```diff
@@ -0,0 +1,24 @@
+/*
```
surovic, line 1: Should be part of `utils.h`

unittests/src/taintdag/tdag.cpp
```diff
@@ -0,0 +1,397 @@
+#include <catch2/catch.hpp>
+
+#include <algorithm>
+#include <cassert>
```
surovic, line 4: Including `cassert` in testing code seems weird. What do we need this for?

surovic (changes requested):

mypy.ini
hbrodin: Was added since we had mypy complain on a lot of files that aren't supported anymore. I did not want to mess around with deleting them and do even more stuff in this PR. I can do a follow-up PR and try to revert these back, if that's OK with you?

polytracker/include/taintdag/util.h
hbrodin: Would you prefer I just delete it a replace usage with more explicit code? I can do that!

polytracker/src/taintdag/polytracker.cpp
hbrodin: I'll fix!

unittests/src/taintdag/tdag.cpp
hbrodin: I think it is a left over from early attempts. I'll remove it.

tests/conftest.py
hbrodin: Correct. Will remove.

polytracker/include/taintdag/util.h
surovic: I would. Alternatively there's [tl::optional](https://github.com/TartanLlama/optional).

unittests/src/taintdag/tdag.cpp
surovic: Working on this already.

unittests/src/taintdag/error_exit_helper.h
surovic: Working on this already.

unittests/src/taintdag/bitmap_section.cpp
surovic: Working on this already.

mypy.ini
surovic: > I can do a follow-up PR and try to revert these back, if that's OK with you?

Sounds good!

polytracker/src/taintdag/polytracker.cpp
surovic: Done.

tests/conftest.py
surovic: Done.

polytracker/src/taintdag/polytracker.cpp
```diff
@@ -7,212 +7,163 @@
...
-    was_reused = range.has_value();
-  }
-
-  if (!range) {
-    auto fsize = details::file_size(fd);
-    if (fsize) {
-      range = tdag_.reserve_source_labels(fsize.value());
+void PolyTracker::open_file(int fd, std::filesystem::path const &path) {
+  // Try to determine the file size. If unknown, just mark it as unknown.
+  uint64_t size = [fd]() -> uint64_t {
```
surovic, line 30: Can we replace the initialization lambda with [std::filesystem::file_size](https://en.cppreference.com/w/cpp/filesystem/file_size) ?

surovic (changes requested):

polytracker/src/taintdag/polytracker.cpp
hbrodin: For sure! Should have been done that way in the first place...

mypy.ini
hbrodin: Added #6507 to ensure we remember.

polytracker/include/taintdag/section.h
```diff
@@ -137,11 +137,13 @@ template <typename T> struct FixedSizeAlloc : SectionBase {
   // N.b. count()/size() competes for the same lock.
   template <typename... Args>
   std::optional<ConstructCtx> construct(Args &&... args) {
-    return map(SectionBase::write(entry_size()), [&](auto &ctx) {
-      return ConstructCtx{.ctx = std::move(ctx),
-                          .t = *new (&*ctx.mem.begin())
+    if (auto write_context = SectionBase::write(entry_size()); write_context) {
```
surovic, line 140:
```C
if (auto write_context = SectionBase::write(entry_size())) {...}
```
Should be sufficient, right? Or am I missing something? Same with other changes in this commit.

surovic (changes requested):

polytracker/include/taintdag/section.h
hbrodin: Good catch! Fixed in 58d650f .

polytracker/mapping.py
hbrodin: @surovic  I believe we can resolve this after our discussion, right?
The same DFS is used as previously. Also a bug was uncovered and is solved by this new implementation. We would incorrectly mark source offsets as used in case of multiple source files. Have probably not been a problem in practice though.

PR #6502, opened 2022-11-09
Adds sections needed for function tracing

surovic:


polytracker/src/taintdag/fntrace.cpp
```diff
@@ -16,35 +16,16 @@ namespace taintdag {
...
-}
-
-void FnTrace::log_fn_event(event_t::kind_t kind, fn_index_t idx) {
-  std::unique_lock write_lock{memory_m};
-  auto maybe_offset{write_event({kind, idx})};
-  if (!maybe_offset) {
-    error_exit("Failed to log event with id: ", get_event_count());
+void Events::log_fn_event(Event::kind_t kind, Functions::index_t idx) {
+  // Write an `Event` via `construct`
+  auto maybe_ctx{construct(kind, idx)};
```
hbrodin, line 25: The `maybe_ctx` isn't really used, could just `if (!construct(kind, idx)) {` if you want to.

polytracker/taint_dag.py
```diff
@@ -314,6 +338,7 @@ def __init__(self, file: BinaryIO) -> None:
         self.sink_cache: Dict[int, TDSink] = {}
 
         self.fd_headers: List[Tuple[Path, TDFDHeader]] = list(self.read_fd_headers())
+        self.fn_headers: List[Tuple[str, TDFnHeader]] = list(self.read_fn_headers())
```
hbrodin, line 341: I guess this could produce some overhead on large traces (with many functions). But, maybe it isn't that many. We could always change it to be lazy at some later point if needed. The reasoning for `fd_headers` is probably that there aren't that many files being handled in a TDAG file.
Anyhow, for now I think it is good that they follow the same pattern.

tests/test_fntrace.py
```diff
@@ -1,28 +1,25 @@
...
-#     assert kinds[taint_dag.TDEvent.Kind.ENTRY] == kinds[taint_dag.TDEvent.Kind.EXIT]
+@pytest.mark.program_trace("test_fntrace.cpp")
+def test_fntrace(program_trace: ProgramTrace):
+    assert isinstance(program_trace, taint_dag.TDProgramTrace)
+    events = list(program_trace.tdfile.events)
+    assert len(events) == 10
+    kinds: Dict[taint_dag.TDEvent.Kind, int] = defaultdict(int)
+    for e in events:
+        kinds[e.kind] += 1
+    assert kinds[taint_dag.TDEvent.Kind.ENTRY] == kinds[taint_dag.TDEvent.Kind.EXIT]
```
hbrodin, line 25: Clever way of checking it!

unittests/src/taintdag/fnmapping.cpp
```diff
@@ -10,32 +10,31 @@
...
-    fnm.add_mapping("bar");
-    fnm.add_mapping("baz");
-    auto foo_2{fnm.add_mapping("foo").value_or(4)};
+    td::OutputFile<td::StringTable, td::Functions> of{std::tmpnam(nullptr)};
+    auto &functions{of.section<td::Functions>()};
+    auto foo_1{functions.add_mapping("foo").value_or(3)};
+    functions.add_mapping("bar");
+    functions.add_mapping("baz");
+    auto foo_2{functions.add_mapping("foo").value_or(4)};
     REQUIRE(foo_1 == foo_2);
```
hbrodin, line 38: Does this test pass? I can't figure out how it would run successfully. There is nothing checking for duplicates unless I'm totally misunderstanding something (which, of course is also an option 😄 ).

unittests/src/taintdag/fntrace.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+TEST_CASE("Test fntrace operations") {
+  namespace td = taintdag;
+  SECTION("Log unique events") {
+    td::OutputFile<td::Events> of{std::tmpnam(nullptr)};
+    auto &events{of.section<td::Events>()};
+    td::Functions::index_t fnidx{0};
+    events.log_fn_event(td::Event::kind_t::entry, fnidx);
+    events.log_fn_event(td::Event::kind_t::exit, fnidx);
+    SECTION("Events are successfully written") { REQUIRE(events.count() == 2); }
```
hbrodin, line 21: Maybe a check to ensure that `fnidx` is actually stored?

unittests/src/taintdag/utils.h
```diff
@@ -0,0 +1,94 @@
+/*
```
hbrodin, line 1: Good move renaming and cleaning this file!

hbrodin (commented):
Nice! I think this is almost ready to merge. There are a couple of things I would prefer we look into. First, the test case for unique names looks suspicious to me. I don't think we run unittests as part of CI, that is something I think we should do (not in this PR). Then, there are a couple of minor things to check (see comments below).

Dockerfile
```diff
@@ -14,7 +14,9 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get -y update  \
       gdb                                             \
       libncurses5-dev                                 \
       apt-transport-https                             \
-      ca-certificates
+      ca-certificates                                 \
+      libstdc++-10-dev
```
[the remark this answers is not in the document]
surovic: To get `concepts` into the `opt` pass plugin. The pass plugin is compiled with the system `stdlib` instead of the instrumented one we provide. It makes sense, since we don't really need to link the LLVM pass plugin against instrumented code.

polytracker/include/taintdag/fnmapping.h
```diff
@@ -9,74 +9,40 @@
...
-  std::optional<offset_t> write_name(std::string_view name);
-  std::optional<offset_t> write_header(header_t header);
-  // Cache
-  std::unordered_map<std::string_view, index_t> mappings;
-};
-
-// TODO (hbrodin): Unsure what goes in the FunctionEntry atm.
-struct FunctionEntry {};
+  static constexpr uint8_t tag{6};
+  static constexpr size_t allocation_size{std::numeric_limits<index_t>::max() *
```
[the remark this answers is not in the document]
surovic: I think you're right. I've used [`Sources`](https://github.com/trailofbits/polytracker/blob/new_fn_mapping/polytracker/include/taintdag/taint_source.h#L59-L60) as reference. Does this mean that the code in `Sources` should also be modified? Or is there a reason why there's not a `+1` there?

polytracker/taint_dag.py
surovic: Agreed.

polytracker/include/taintdag/fnmapping.h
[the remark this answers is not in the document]
hbrodin: Ohh... Yeah, I believe there should be a +1 there as well. I believe nothing bad will happen, we just won't be able to store the last item. There will be a limit check in the `construct`-method to prevent out of bounds writes.

polytracker/include/taintdag/fntrace.h
```diff
@@ -11,34 +11,24 @@
...
-  std::optional<offset_t> write_event(event_t event);
+  enum class kind_t : uint8_t { entry, exit };
+  kind_t kind;
+  Functions::index_t function;
+};
+
+class Events : public FixedSizeAlloc<Event> {
+public:
+  static constexpr uint8_t tag{7};
+  static constexpr size_t allocation_size{std::numeric_limits<uint32_t>::max() *
```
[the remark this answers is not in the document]
surovic: In the end I decided not to do the `+1` anywhere because in the `Events` section case it overflows `allocation_size`. The `+1` is also unnecessarily confusing and overall not worth the one extra entry in any section.

unittests/src/taintdag/fntrace.cpp
hbrodin: I think the way you did it is perfectly fine. We could consider other interfaces/APIs going forward, but for now I believe this is just fine.

hbrodin (commented):
We need to figure out a way to synchronize access to the mapping cache, besides that it is good to merge!

hbrodin (approved):
LGTM

PR #6508, opened 2022-11-18
examples: DaeDaLus nitf parser demo

kaoudis:


polytracker/src/passes/bb_splitting_pass.cpp
```diff
@@ -0,0 +1,120 @@
+/*
```
josephsweeney, line 1: I think we can revert this, there was a one line change that I originally needed for M1 compatibility and I think this file is just no longer relevant. I can make the change and test if you want.

kaoudis: sure, go for it! thank you

josephsweeney: Done. I couldn't find the PR that remove that file though, I thought it must have been the sections PR but I don't see it. 🤔 Probably best to wait for CI before a merge.

josephsweeney (approved):

ESultanik (approved):

PR #6512, opened 2022-11-21
Adds sockets as taint sinks

### LINKED ISSUE #6489: Add support for tracking writes to sockets, similar to how we handle STDOUT

hbrodin:
This implements support for AF_INET sockets to be used as taint sinks. Whenever tainted output is written to a socket it is recorded in the taint log.

Fixes #6489.


polytracker/include/taintdag/stream_offset.h
```diff
@@ -26,10 +26,11 @@ template <size_t SourceCount> class StreamOffset {
   // concurrent read operations for the same source is the result of invalid
   // code. Reads on the same source should be protected from running in
   // parallell (at least for the taint sources considered here, e.g. files).
-  source_offset_t read(source_index_t idx, size_t len) {
+  source_offset_t increase(source_index_t idx, size_t len) {
```
kaoudis, line 29: If even after the changes here this function is about updating the cursor position on the source, it seems like based on the way this function is used `increase` is a more accurate way of thinking about the operation? Is there ever a case where we might loop or recurse through input, or otherwise accept a decreasing idx compared to wherever our last position was?

kaoudis (commented):
Overall this seems fairly straightforward and legit to me, but knowing I may lack context Marek or Evan have, please take that with a grain of salt :)

A minor general question: the use of sockets seems to imply we could write out of the container to a remote address if we wanted to? Do you plan to remotely write polytracker output to (not sure) make further operations over the output easier?

polytracker/include/taintdag/stream_offset.h
kaoudis: If this is more about cases where we don't have the ability to retrieve offset otherwise (from https://github.com/trailofbits/polytracker/pull/6512/files#diff-07ef4fa57e851139977d8f9436d354eee69daf7d3be3b551aaa351ba246fe9faR143 it seemed we always wrap offset usage and wouldn't query the os directly, but it seems like that was a mistaken impression), that makes sense!

hbrodin: We have two general cases when determining the offset:
1. Regular files etc, things we can successfully do `lseek` on to get the current offset
2. Sockets, pipes etc, things where `lseek` returns an error.

If we detect case 2, the `increase` function is used for every call to e.g. `write` on a socket, starting from zero and increasing by the number of written bytes. We track the offset of the output separately, in a StreamOffset instance, and assume it started from zero.

The same thing happens on e.g. reads from sockets, it works exactly the same, except that we increase the offset by the amount read.

I don't know if the "wrapped offset usage" you refer to is the usage of `util::Offset`? That class attempts to get the offset via `lseek`, and indicates if it was successful or not.

kaoudis: > I don't know if the "wrapped offset usage" you refer to is the usage of `util::Offset`
yep! til now I had only taken a *very* quick look at util::Offset to see if it was polytracker code or not since util.cpp wasn't otherwise mentioned in this PR, so wasn't 100% sure what all was in there beyond "it exists", apologies

Thanks for explaining, I appreciate it :D

ESultanik (approved):
Looks good to me!

PR #6513, opened 2022-11-21
Remove exclusion filters from mypy.ini as it is not needed anymore

### LINKED ISSUE #6507: Investigate mypy supressions
PR #6500 adds suppressions to the mypy.ini file. This is a reminder to look into if we can remove those suppressions.

hbrodin:
Only include suppressions and configuration for that is currently needed.

Fixes #6507


kaoudis (commented):
Does   https://github.com/trailofbits/polytracker/blob/master/.github/workflows/dockerimage.yml somehow use this ini file or does this only apply to local runs? I think this is the relevant line of the Action job :
```
mypy --python-version ${{ matrix.python-version }} --ignore-missing-imports polytracker tests
```

mypy.ini
```diff
@@ -1,13 +1,2 @@
...
-exclude = (?x) (
-  diffing.py$ |
-  polytracker/grammars.py$ |
-  test_bb_splitting.py$ |
-  tests/test_grammar_extraction_python.py$ |
-  tests/test_grammar_extraction_native.py$
-  )
-
-[mypy-polytracker.mimid.*]
-# Mimid upstream has lots of type errors, so ignore them
```
kaoudis, line 12: The only place I see Mimid mentioned is here - https://github.com/trailofbits/polytracker/search?q=mimid . Was it a prior dep?

kaoudis (approved):

hbrodin:
> Does https://github.com/trailofbits/polytracker/blob/master/.github/workflows/dockerimage.yml somehow use this ini file or does this only apply to local runs? I think this is the relevant line of the Action job :
> 
> ```
> mypy --python-version ${{ matrix.python-version }} --ignore-missing-imports polytracker tests
> ```

My understanding is that it picks up mypy.ini from the dir and use it for configuration.

kaoudis:
> My understanding is that it picks up mypy.ini from the dir and use it for configuration.

yeah that makes sense - looks like it implicitly picks it up by default https://mypy.readthedocs.io/en/stable/config_file.html but one can override with `--config-file`, just not required by default

mypy.ini
hbrodin: Yes, I believe it was related to the grammar recovery previously attempted. It is not used anymore.

PR #6517, opened 2022-11-22
Ports over dfsan from `polytracker-llvm`

surovic:
The ports over the `DataflowSanitizer` pass and related parts of `compiler-rt` from our modified fork of the LLVM repository. By doing this we are no longer dependent on that fork and can eventually move to more recent versions of LLVM. For this to happen, parts of polytracker's CMake build system had to be modified and the Dockerfile had to be rewritten. Users should also expect increased docker image build times as we now build `libc++`, which was part of the `polytracker-llvm` docker dependency build. The sum duration of `polytracker-llvm` and `polytracker` docker image build has however been drastically lowered. Future hope is to optimize the docker image build further.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -133,7 +133,6 @@ ENV DFSAN_OPTIONS="strict_data_dependencies=0"
 ENV WLLVM_BC_STORE=/project_bitcode
 ENV WLLVM_ARTIFACT_STORE=/project_artifacts
 
-RUN mkdir $WLLVM_ARTIFACT_STORE
-RUN mkdir $WLLVM_BC_STORE
+RUN mkdir $WLLVM_ARTIFACT_STORE && mkdir $WLLVM_BC_STORE
 
 ENV PATH=$PATH:/polytracker-install/bin


Dockerfile
```diff
@@ -1,56 +1,129 @@
...
-WORKDIR /polytracker/build
-RUN cmake -GNinja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_VERBOSE_MAKEFILE=TRUE -DCXX_LIB_PATH=/cxx_libs ..
-RUN ninja install
+RUN pip3 install /polytracker
 
+FROM polytracker-python as polytracker-cxx
 
-WORKDIR /polytracker
-RUN pip3 install .
+ARG DFSAN_FILENAME_ARCH=x86_64
```
hbrodin, line 103: Are we able to build this on other archs? We used to have some (limited) support for arm-builds, I think.

polytracker/include/polytracker/passes/DataFlowSanitizer.h
```diff
@@ -0,0 +1,31 @@
+//===- DataFlowSanitizer.h - dynamic data flow analysis -------------------===//
```
hbrodin, line 1: This is a modified copy of the llvm-file. Is this the appropriate way of adding copyright notices etc. Should we note that this is a modified version? I'm assuming we should be keeping this copyright header, but maybe add some additional comments.

polytracker/src/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-find_package(LLVM 13 CONFIG REQUIRED)
+find_package(LLVM 12 CONFIG REQUIRED)
```
hbrodin, line 1: OK, so we are using LLVM-12. I think you mentioned why at some point, but I can't remember. Was there some technical difficulty in using 13+?

hbrodin (commented):
Wow! I'm impressed! This is some really extensive work. I have made some comments I'd like to discuss. Also, some general notes:
* Is it recorded somewhere what version of the compiler-rt that is included here? I think it would be good to add that information somewhere.
* I've assumed that anything in `compiler-rt` is untouched so I haven't paid to much attention to those files. Is that correct?

Dockerfile
```diff
@@ -1,56 +1,129 @@
...
+  clang-12                                            \
+  cmake                                               \
+  git                                                 \
+  file
+
+RUN update-alternatives --install /usr/bin/opt opt /usr/bin/opt-12 10
+RUN update-alternatives --install /usr/bin/llvm-link llvm-link /usr/bin/llvm-link-12 10
+RUN update-alternatives --install /usr/bin/llvm-ar llvm-ar /usr/bin/llvm-ar-12 10
+RUN update-alternatives --install /usr/bin/clang clang /usr/bin/clang-12 10
+RUN update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-12 10
```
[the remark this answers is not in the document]
surovic: >Does this mean we are using llvm-12? Aren't we using llvm-13 in the current branch?

We're installing `clang` and related bitcode utilities via `apt` in version 12 because the `clang-13` package contains version `13.0.1` which has weird behavior wrt `-Werror`. The flag seems to be raised even if the user doesn't explicitly specify so. I recall seeing some discussion about this in official LLVM channels and it was done to mimic `gcc` behavior. `MuPDF` doesn't build with `clang-13` installed from `apt` for example. The version we use in `polytracker-llvm` is `13.0.0`.

> It seems from the clone below that we build libcxx from version 13. How does this work?

The version mismatch here doesn't really matter in my opinion. We're building `libcxx` using `clang-12`.

Dockerfile
```diff
@@ -1,56 +1,129 @@
...
+  -DCMAKE_INSTALL_PREFIX=$LIBCXX_INSTALL_DIR \
+  -DLIBCXXABI_ENABLE_SHARED=NO \
+  -DLIBCXX_ENABLE_SHARED=NO \
+  -DLIBCXX_CXX_ABI="libcxxabi" \
+  -DLLVM_ENABLE_LIBCXX=ON \
+  -DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi"
+
+RUN cmake --build $LIBCXX_BUILD_DIR --target install-cxx install-cxxabi -j$((`nproc`+1))
+
+FROM clean-libcxx as polytracker-libcxx
```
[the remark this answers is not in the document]
surovic: Actually, I don't think these do anything besides provide targets to specify via `docker build`. But it can be a start of something!

Dockerfile
```diff
@@ -1,56 +1,129 @@
...
+ENV PATH="$PATH:/root/go/bin"
+
+FROM base as clean-libcxx
+
+ENV BITCODE=/cxx_clean_bitcode
+RUN mkdir -p $BITCODE
+ENV WLLVM_BC_STORE=$BITCODE
+
+ENV CC="gclang"
+ENV CXX="gclang++"
```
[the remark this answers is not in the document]
surovic: It was part of it, yes. The last issue was resolved via using gclang built libcxx  [here](https://github.com/trailofbits/polytracker/blob/58401fb97b71c79861ab8a0179e5d9d811547b06/Dockerfile#L111).

Dockerfile
```diff
@@ -1,56 +1,129 @@
...
+  -DCMAKE_INSTALL_PREFIX=$LIBCXX_INSTALL_DIR \
+  -DLIBCXX_ABI_NAMESPACE="__p" \
+  -DLIBCXXABI_ENABLE_SHARED=NO \
+  -DLIBCXX_ENABLE_SHARED=NO \
+  -DLIBCXX_ABI_VERSION=2 \
+  -DLIBCXX_CXX_ABI="libcxxabi" \
+  -DLIBCXX_HERMETIC_STATIC_LIBRARY=ON \
+  -DLIBCXX_ENABLE_STATIC_ABI_LIBRARY=ON \
+  -DLLVM_ENABLE_LIBCXX=ON \
+  -DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi"
```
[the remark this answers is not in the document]
surovic: Honestly, I have no idea either. I see no reason why we should need two builds of `libcxx`, however removing either of those builds results in `libcxx` symbols missing from the final instrumented binary. It's one of the first things I hope to investigate when I get back to these issues.

Dockerfile
surovic: I don't see a reason why we wouldn't be able to. That being said, I fully expect there to be issues.

polytracker/include/polytracker/passes/DataFlowSanitizer.h
surovic: I don't really have an opinion on this. Maybe @ESultanik could chime in?

polytracker/src/CMakeLists.txt
surovic: See comments above.

polytracker/src/compiler-rt/include/CMakeLists.txt
```diff
@@ -0,0 +1,104 @@
...
+  set(PROFILE_HEADERS
+    profile/InstrProfData.inc
+    )
+endif(COMPILER_RT_BUILD_PROFILE)
+
+set(COMPILER_RT_HEADERS
+  ${SANITIZER_HEADERS}
+  # ${FUZZER_HEADERS}
+  # ${MEMPROF_HEADERS}
+  # ${XRAY_HEADERS}
```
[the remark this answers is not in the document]
surovic: I'm not sure. Since we built a modified version of the whole `compiler-rt` previously, we did have XRAY included in it. The `compiler-rt` excerpt from this PR doesn't have it included. Theoretically it should be possible to link against an `apt` installed version of `compiler-rt` but I haven't tried that.

Dockerfile
surovic: Thinking about this a little bit more. Do we have use-cases where we want to build for say ARM on an X86 host? Cause if we omit the target argument altogether it will by default build for the host arch.

Dockerfile
```diff
@@ -1,56 +1,129 @@
...
-            vim                                             \
-            file                                            \
-            gdb                                             \
-            libncurses5-dev                                 \
-            apt-transport-https                             \
-            ca-certificates                                 \
-            libstdc++-10-dev
+
+ARG BUILD_TYPE="Release"
+ARG PARALLEL_LINK_JOBS=1
```
ESultanik, line 6: Any reason why this is set to `1`? Are there race conditions during linking?

Dockerfile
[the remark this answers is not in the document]
ESultanik: I agree that the version mismatch doesn't really matter in this case. It would be nice to copy/paste this reasoning as a comment, though.
```suggestion
RUN update-alternatives --install /usr/bin/opt opt /usr/bin/opt-12 10
RUN update-alternatives --install /usr/bin/llvm-link llvm-link /usr/bin/llvm-link-12 10
RUN update-alternatives --install /usr/bin/llvm-ar llvm-ar /usr/bin/llvm-ar-12 10
RUN update-alternatives --install /usr/bin/clang clang /usr/bin/clang-12 10
RUN update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-12 10
# We install clang and related bitcode utilities via apt in version 12 because the clang-13 package contains version 13.0.1 which has weird behavior wrt -Werror.
# The flag seems to be raised even if the user doesn't explicitly specify so.
# We believe this is intentional on the part of LLVM to mimic gcc behavior.
# MuPDF doesn't build with clang-13 installed from apt, for example.
```

polytracker/include/polytracker/passes/DataFlowSanitizer.h
ESultanik: We could consider adding the git commit hash of the version from which we forked, but I think this is otherwise fine. We might want to update our base README with copyright notices from forked/vendored LLVM code if we haven't already done so.

ESultanik (approved):

Dockerfile
```diff
@@ -1,56 +1,129 @@
...
+RUN DEBIAN_FRONTEND=noninteractive apt-get -y update
+RUN DEBIAN_FRONTEND=noninteractive apt-get -y install \
+  ninja-build                                         \
+  python3-pip                                         \
+  python3.8-dev                                       \
+  python-is-python3                                   \
+  ca-certificates                                     \
+  libstdc++-10-dev                                    \
+  golang                                              \
+  clang-12                                            \
```
kaoudis, line 18: Should we still depend on / require Clang 12 if the goal is to move away from old Clang? Or are there plans to update this version-by-version once the removing-polytracker-llvm change lands?

Dockerfile
```diff
@@ -1,56 +1,129 @@
...
-            apt-transport-https                             \
-            ca-certificates                                 \
-            libstdc++-10-dev
+
+ARG BUILD_TYPE="Release"
+ARG PARALLEL_LINK_JOBS=1
+ARG LLVM_TARGET_NAME=X86
+
+RUN DEBIAN_FRONTEND=noninteractive apt-get -y update
+RUN DEBIAN_FRONTEND=noninteractive apt-get -y install \
```
kaoudis, line 10: Since each `RUN` command creates a new Docker ephemeral layer, prefer combining RUN commands once the ordering and commands themselves are relatively the way you want them, e.g.:

```
    RUN apt-get -y update && \
             apt-get -y install \
                      ninja-build \ 
                      ...
```

This minimises the number of ephemeral containers which need to be created to run the Dockerfile.

Dockerfile
```diff
@@ -1,56 +1,129 @@
...
-            libncurses5-dev                                 \
-            apt-transport-https                             \
-            ca-certificates                                 \
-            libstdc++-10-dev
+
+ARG BUILD_TYPE="Release"
+ARG PARALLEL_LINK_JOBS=1
+ARG LLVM_TARGET_NAME=X86
+
+RUN DEBIAN_FRONTEND=noninteractive apt-get -y update
```
kaoudis, line 9: Prefer setting `DEBIAN_FRONTEND` in the Dockerfile environment once prior to all commands that need it, e.g. 

```
ENV DEBIAN_FRONTEND=noninteractive
```
before all `RUN` commands where you prefix setting `DEBIAN_FRONTEND`.

Dockerfile
```diff
@@ -1,56 +1,129 @@
...
+  python-is-python3                                   \
+  ca-certificates                                     \
+  libstdc++-10-dev                                    \
+  golang                                              \
+  clang-12                                            \
+  cmake                                               \
+  git                                                 \
+  file
+
+RUN update-alternatives --install /usr/bin/opt opt /usr/bin/opt-12 10
```
kaoudis, line 23: Can all these `update-alternatives` calls be combined into a single call? If not, they can at least be `&&`ed together into a single `RUN` command as noted above.

Dockerfile
```diff
@@ -1,56 +1,129 @@
...
-            curl && \
-      # need a pip compatible with 3.10
-      curl -sS https://bootstrap.pypa.io/get-pip.py | python3.10 && \
-      update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.10 10 && \
-      python3 -m pip install pytest
-
-WORKDIR /blight
-RUN git clone https://github.com/trailofbits/blight.git .
-RUN pip3 install .
+RUN GO111MODULE=off go get github.com/SRI-CSL/gllvm/cmd/...
```
kaoudis, line 37: Thanks for updating the gllvm! I may have missed it, but what golang version is supported in this new Dockerfile version? I think the current `master` Polytracker base Dockerfile is pinned to an old Golang as well.

Dockerfile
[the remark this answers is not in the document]
kaoudis: Perhaps the prior-stage libcxx-abi build (and symbols) may no longer be accessible or are not accessible where expected at final-binary build time, due to the multi-stage nature of the build? This is an untested hypothesis, please take it with a grain or two of salt 😅

Dockerfile
kaoudis: Iirc, @josephsweeney would like to build on Apple Silicon instead of spinning up a cloud Linux VM or similar. But I don't think that should require ARM builds from an x86 host!

Dockerfile
```diff
@@ -1,56 +1,129 @@
...
+RUN git clone --depth 1 --branch llvmorg-13.0.0 https://github.com/llvm/llvm-project.git
+RUN git clone --depth 1 --branch master https://github.com/trailofbits/blight.git
+
+RUN pip3 install pytest /blight
 
 RUN update-ca-certificates
 
-RUN apt-get install -y software-properties-common && \
-      # deadsnakes PPA allows us to install python3.10 on focal
-      add-apt-repository ppa:deadsnakes/ppa && \
```
kaoudis, line 25: thanks for removing this, I was hoping to get around to it once we were fully away from focal, so this is perfect

Dockerfile
hbrodin: IIRC correctly it is really heavy on memory and having higher values might cause OOM-death.

Dockerfile
josephsweeney: Yeah, I don't expect any cross-architecture build necessary. Just making it possible to build on both x86 and arm64 would be good. I also don't think this has to happen immediately, we can get an arm build working when needed.

Dockerfile
surovic: The argument was removed as it was unused.

Dockerfile
```diff
@@ -1,56 +1,129 @@
-FROM trailofbits/polytracker-llvm:bbdb72d136af011db9710a4d44a79e6d5059cc1a
+FROM ubuntu:jammy as base
```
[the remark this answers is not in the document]
surovic: I've added comments explaining the build steps. Let me know what you think!

Dockerfile
surovic: Done! Hopefully merged all `RUN` commands that could be merged.

Dockerfile
surovic: Done!

Dockerfile
surovic: Done!

Dockerfile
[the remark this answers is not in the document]
surovic: Added a comment!

surovic:
> Is it recorded somewhere what version of the compiler-rt that is included here? I think it would be good to add that information somewhere.

I'll make an issue to add a note in a new `polytracker/src/compiler-rt/README.md`.

> I've assumed that anything in `compiler-rt` is untouched so I haven't paid to much attention to those files. Is that correct?

Yes. That is correct.

PR #6520, opened 2022-11-30
github: run a given actions lint only if files of the right type change (and lint the files)

kaoudis:
In order to run GH Actions jobs only on specific file types in the cleanest way I could come up with (specifying file type paths to run on), split up the `basic_ci` job into several. While there are some minorly duplicated steps between the jobs, if we are very concerned about this, it is possible to cache outputs of job1 and feed into job2. However, since job1 and job2 might not even need to run on the same commit (a commit not including c++ files will not trigger `lint-cxx`, for example), I chose to keep it simple(r) for now.

Ancillary changes to help things along:
- use actions/checkout's `submodules: recursive` option instead of `submodules: true`, since as in https://github.com/actions/checkout/issues/51 we do not get the submodules checked out otherwise, which seems to result in the clang linter's base py file not existing
- expand lint coverage within the repo as much as reasonably possible
- add shellcheck + shfmt (`lint-shell`)
- make a couple minor changes [1](https://github.com/trailofbits/polytracker/compare/master...kaoudis/gh-actions-linting#diff-0236c76d004cd073861385152aa4afe2a4d02909d52d24b392cf880dd1f8e9e8R46), [2](https://github.com/trailofbits/polytracker/compare/master...kaoudis/gh-actions-linting#diff-8c5c8909a923331b4cec18132468b7ac16b8aebe98efd0e88ebbd8cf869ca217R6) beyond what I would consider "lint" or "format" modifications so that mypy doesn't get mad when run with python3.7 (not sure why we weren't getting the errors before)


.github/workflows/cxx.yml
```diff
@@ -0,0 +1,36 @@
+name: 🥥 Lint C++
+
+on:
+  workflow_dispatch: {}
+  push:
+    paths:
+    - "**.cpp"
+    - "**.cc"
```
hbrodin, line 8: Suggest we either add to the 'run'-list below or remove '.cc'/'.hpp' here to ensure they react on the same extensions.

.github/workflows/shell.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Shellcheck and shformat
+        uses: luizm/action-sh-checker@master
+        with:
+          sh_checker_exclude: "third_party/"
+          sh_checker_comment: true
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
hbrodin, line 20: I'm probably missing something, but why do we need the `GITHUB_TOKEN` here?

hbrodin (commented):
This is great! I just made a couple of minor comments. The three `basic_ci`-jobs, seems to be stuck. Not sure if that is intentional. Also, we need to make sure it can be merged.
It seems to be quite a lot of changes, almost all of them are formatting changes if I'm not mistaken.

ESultanik (approved):
LGTM, but I'd suggest we remove the `GITHUB_TOKEN` thing that @hbrodin mentioned or, if it's not possible, determine why the third party action requires it.

kaoudis:
> The three `basic_ci`-jobs, seems to be stuck. Not sure if that is intentional.
I think the issue here is those three jobs technically no longer exist on this branch. When adding more lints, I split up basic_ci from several jobs in a file into separate Actions since especially with more different kinds of lints added it was doing a lot of disparate stuff, and then removed the empty file which no longer contained any jobs. If that is not cool, I can merge its parts back together! I will take another look at this to be sure that is the cause, but last I recall, that was all that was going on 😅

.github/workflows/cxx.yml
kaoudis: Sure, happy to remove `.cc` and `.hpp` here. [Paths](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths) reference in case it is useful - I added the other c++ extensions on top of what we have in the current working repo copy because I've seen them used in a couple commits, but you're right in that `run-clang-format.py` would never format files with those extensions as configured, since iirc we have to pass specific folders to it... I might try just passing it the root locally on this branch and if that doesn't work, will remove `.cc` and `.hpp` from the `paths` list which triggers this Action.

.github/workflows/shell.yml
kaoudis: Shellcheck wants it in order to comment on the diff, but I don't think that is useful actually, since we don't have anything else that comments on diffs and usually check the Action output. I'll remove it (and the related `sh_checker_comment: true`).

PR #6526, opened 2022-12-13
Return a `TDUntaintedNode` for taint label 0.

### LINKED ISSUE #6524: Erroneous 'RangeNode' is emitted on enumeration
Compiling an instrumented version of the below program
```c++
#include <limits>
#include <iostream>
#include <unistd.h>

int main() {

  int base = std::numeric_limits<int>::max() - 100;
  char c1 = 0, c2 = 0;

  read(0, &c1, sizeof(c1));
  if ((c1 & 1) == 0)
  {
    printf("c1 is even %x\n", c1);
  }
  else
  {
    printf("c1 is odd %x\n", c1);
    read(0, &c2, sizeof(c2));
    if (c2 + base < base) {
      printf( "c2 was to big for base! got %x", (c2 + base));
    } else {
      printf("Perfect! C: %x", c2);
    }

  }
  return 0;
}
```

Yields the following node sequence, when iterating the resulting TDAG:
```shell
TDRangeNode: affects control flow False [0, 0]
TDSourceNode: affects control flow True idx 1 offset 0
TDSourceNode: affects control flow True idx 1 offset 1
```

using
```python
tr = PolyTrackerTrace.load("tdagfile")
for n in tr.tdfile.nodes:
    print(f"{n}")
```

I think the reason is that label zero is interpreted as a RangeNode when it should be omitted.

hbrodin:
Fixes #6524

### DIFF
--- a/polytracker/taint_dag.py
+++ b/polytracker/taint_dag.py
@@ -262,6 +262,14 @@ class TDUnionNode(TDNode):
         return f"TDUnionNode: {super().__repr__()} ({self.left}, {self.right})"
 
 
+class TDUntaintedNode(TDNode):
+    def __init__(self):
+        super().__init__(False)
+
+    def __repr__(self) -> str:
+        return f"TDUntaintedNode: {super().__repr__()}"
+
+
 class TDSink(Structure):
     """Python representation of the SinkLogEntry from sink.h"""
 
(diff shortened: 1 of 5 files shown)


polytracker/taint_dag.py
```diff
@@ -389,11 +397,13 @@ def decode_node(self, label: int) -> TDNode:
             if v1 > v2:
                 return TDUnionNode(v1, v2, affects_cf)
             else:
+                if v == 0:
```
surovic, line 400: nitpick: probably better to move this one nest up and do
```python
if v == 0:
...
elif v1 > v2:
...
else:
...
```

surovic (approved):

polytracker/taint_dag.py
hbrodin: My thinking is that this should be a rare case, that's why I didn't want to put it at the top. But maybe it should go there instead? Something like:
```python
if label == 0:
    return TDUntaintedNode()
````

surovic: Yeah, if that 's possible, I think we should do that! Just to get the obvious cases out the way ASAP in the code,

Issue #6528, opened 2022-12-19
Applications that exit with `assert` kills the `polytracker info` command


hbrodin:
If an application exits due to an assert (or otherwise unexpected exit). The size fields for each section remains at it's [allocated size](https://github.com/trailofbits/polytracker/blob/master/polytracker/include/taintdag/outputfile.h#L100). Under normal circumstances, [that size is updated when the application exits](https://github.com/trailofbits/polytracker/blob/master/polytracker/include/taintdag/outputfile.h#L119-L131).

A first mitigation to prevent crash and just return an empty output file could be to initialize each section size to zero. Other options to explore is to have each section update the size as soon as anything is written (requiring an additional write for each write to a section).

PR #6529, opened 2023-01-03
Basic support for assert()

hbrodin:
Will intercept the __assert_fail call and call polytracker_end which will update the output file and close it.
At least a partial solution to #6528.

### DIFF
--- a/polytracker/custom_abi/dfsan_abilist.txt
+++ b/polytracker/custom_abi/dfsan_abilist.txt
@@ -154,7 +154,8 @@ fun:llrintl=functional
 
 # Functions that produce an output that does not depend on the input (shadow is
 # zeroed automatically).
-fun:__assert_fail=discard
+fun:__assert_fail=uninstrumented
+fun:__assert_fail=custom
 fun:__ctype_b_loc=discard
 fun:__cxa_atexit=discard
 fun:__errno_location=discard
@@ -547,7 +548,6 @@ fun:__asinl_finite=uninstrumented
 fun:__asprintf=uninstrumented
 fun:__asprintf_chk=uninstrumented
 fun:__assert=uninstrumented
-fun:__assert_fail=uninstrumented
 fun:__assert_perror_fail=uninstrumented
 fun:__atan2_finite=uninstrumented
 fun:__atan2f_finite=uninstrumented
--- a/polytracker/include/taintdag/outputfile.h
+++ b/polytracker/include/taintdag/outputfile.h
@@ -88,7 +88,7 @@ public:
 
   // The FileHeader is constructed at offset zero of the mapped memory (file).
   // It contains the metadata required to parse a TDAG file. Initially each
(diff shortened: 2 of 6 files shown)


polytracker/custom_abi/dfsan_abilist.txt
```diff
@@ -154,7 +154,8 @@ fun:llrintl=functional
 
 # Functions that produce an output that does not depend on the input (shadow is
 # zeroed automatically).
-fun:__assert_fail=discard
+fun:__assert_fail=uninstrumented
```
kaoudis, line 157: What does "uninstrumented" result in? Does this mean the function is stubbed and polytracker does not interact with it?

polytracker/include/taintdag/outputfile.h
```diff
@@ -97,7 +97,7 @@ template <Section... Sections> class OutputFile {
         (SectionMeta{.tag = Sections::tag,
                      .align = Sections::align_of,
                      .offset = 0,
-                     .size = Sections::allocation_size})...};
+                     .size = 0})...};
```
kaoudis, line 100: Does setting the size to 0 for a section mean the section technically won't exist when output?

polytracker/custom_abi/dfsan_abilist.txt
surovic: >The pass treats every function in the uninstrumented category in the ABI list file as conforming to the native ABI. Unless the ABI list contains additional categories for those functions, a call to one of those functions will produce a warning message, as the labelling behavior of the function is unknown.

Source: https://clang.llvm.org/docs/DataFlowSanitizer.html#abi-list

Does this answer your question?

polytracker/include/taintdag/outputfile.h
surovic: I don't think so? I guess this was done to have correct information in the case when we have a section that's optional, like `Events` if function tracing is turned off. I think there should be a header entry for the section, but it should tell you there's no trace events in it and it's size is 0? Am I right @hbrodin ?

surovic: Also we should update the comment above this code to reflect this change :)

surovic (changes requested):
Just updating the code comment.

polytracker/custom_abi/dfsan_abilist.txt
kaoudis: Thank you! I didn't realize this is/was the same as the regular DFSan ABI list, my bad.

polytracker/include/taintdag/outputfile.h
hbrodin: This is just the initial value. The final value will be set in the destructor of the `OutputFile`. That behaviour hasn't changed, it is just that we don't assume the size to be max from the beginning, but rather to be empty.

kaoudis (approved):
looks like the current changelist made @surovic's requested updates, and I have no further questions!

PR #6537, opened 2023-01-25
Comments out non-functional `polytracker` commands, removes obsoleted package dependecies

surovic:
PR does as title suggests. The commented out commands might prove useful in the future, but pollute public facing API. The removed python package dependencies are AFAIK completely obsolete.

Merge this PR **after** #6535


polytracker/grammars.py
```diff
@@ -44,6 +44,8 @@
     TraceEvent,
 )
 
+from . import PolyTrackerTrace
```
hbrodin, line 47: Why do we need this?

hbrodin (approved):
LGTM! Just a single question. I think it is good to merge (unless my question sheds light on something that needs action).

PR #6552, opened 2023-05-15
Migrates and briefly documents the UBet analysis code for public reference in the LangSec slides. 

kaoudis:
❗ caution: depends on https://github.com/trailofbits/polytracker/pull/6553 and will not accurately reproduce our results without merging that code first / too.


examples/analysis/ubet/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+NITF is a binary image file format. Each NITF packages one or more visual data representations (video, fingerprints, CAT scan, JPEG, etc.) with extra metadata and other conditionally included information e.g., captions, information for rendering visual redactions, or geo-reference data. Nitro parses multiple mutually incompatible versions of the NITF specification. To simulate the effects of encountering a particular bad input we would like reproduce the effects of in a testing, local, or staging environment we applied Nitro instrumented with UBet to a corpus of 148 valid and known-invalid NITF files.
+
+Unfortunately, as also noted in the tags/ subdirectory README, the input files we used to find bugs in Nitro are part of a corpus we accessed during SafeDocs-related evaluation for PolyTracker. As such, we don't have rights to include them here.
```
hbrodin, line 10: I suggest we leave this section out entirely.

examples/analysis/ubet/compress_tdag.py
```diff
@@ -0,0 +1,85 @@
+from typing import List
```
hbrodin, line 1: I know @ESultanik was planning to integrate this in a separate branch. We should probably synch with him as we had discussion about a potential bug that he probably fixed (`ceil` instead of `round`).

examples/analysis/ubet/eval_nitro.py
```diff
@@ -0,0 +1,590 @@
...
+            if debug_steps < release_steps:
+                print_cols(str(dbg_entry[0]), "", dbg_callstack)
+                dbgidx += 1
+            else:
+                print_cols("", str(rel_entry[0]), rel_callstack)
+                relidx += 1
+
+    return
+
+    # dbg_offset_to_func = {}
```
hbrodin, line 346: There is some commented out code here and there. I'd recommend we just remove it.

examples/analysis/ubet/eval_nitro.py
```diff
@@ -0,0 +1,590 @@
...
+    #       dbg_offset_to_func[stroffset] = {func: 1}
+
+    # # print(dbg_offset_to_func)
+    # for k,v in dbg_offset_to_func.items():
+    #   print(k, v)
+
+    # for (i,d) in enumerate(dbg):
+
+
+def custome():
```
hbrodin, line 366: Is this function used anywhere? If not, I think we can remove it.

examples/analysis/ubet/examples/motivation/listing_1.cpp
```diff
@@ -0,0 +1,31 @@
+/* Consider the following (buggy) toy parser, which accepts at least one
+ * command-line argument.
+ *
+ * With optimizations disabled, Clang / LLVM 15.0.0 run on the code in Listing
+ * 1 will produce a binary that operates consistently dependent on input. With
+ * optimizations enabled, Clang will optimize out the first branch of the
```
hbrodin, line 6:
```suggestion
 * optimizations enabled, Clang will, due to undefined behavior, optimize out the first branch of the
```

examples/analysis/ubet/examples/motivation/listing_1.cpp
```diff
@@ -0,0 +1,31 @@
...
+ * conditional, which invokes undefined behavior; the -O3 optimized program
```
hbrodin, line 7:
```suggestion
 * conditional; the -O3 optimized program always returns zero.
```

examples/analysis/ubet/examples/motivation/listing_1.cpp
```diff
@@ -0,0 +1,31 @@
...
+ * always returns zero.
```
hbrodin, line 8:
```suggestion
```

examples/analysis/ubet/examples/ub.cpp
```diff
@@ -0,0 +1,51 @@
+#include <numeric>
```
hbrodin, line 1: Are we actually using this anywhere? If not, we could consider removing it (and run.sh/runub.sh). Maybe good to keep as a minimalistic exampel?

hbrodin (commented):
Great job adding all this documentation! I've made a couple of notes that we could consider. Also, I'm thinking we might want to synch with a couple of other branches going into main.

examples/analysis/ubet/README.md
kaoudis: this section meaning the NITF section? or the NITF standard section? both?

examples/analysis/ubet/compress_tdag.py
kaoudis: The idea with including here was just to satisfy "if I were to dig into the polytracker codebase, is everything I need to reproduce the results in the ubet paper located there?" for a particular point-in-time! I will leave it and let @ESultanik delete this version when he has integrated it into the main polytracker python then.

examples/analysis/ubet/eval_nitro.py
kaoudis: Can do

examples/analysis/ubet/examples/ub.cpp
kaoudis: I was keeping them for purposes of having a minimal example, but we didn't end up using them for more than a learning step iirc, so will remove.

examples/analysis/ubet/Dockerfile.nitro
```diff
@@ -0,0 +1,61 @@
...
+ 	-DCMAKE_CXX_FLAGS="-w -D_POSIX_C_SOURCE=200809L -DCODA_OSS_NO_is_trivially_copyable" \
+   -DCMAKE_BUILD_TYPE=Release \
+	 -DCODA_BUILD_TESTS=OFF \
+	 -DENABLE_PYTHON=OFF
+
+RUN polytracker build cmake --build . -j$((`nproc`+1)) --target show_nitf++ --config Release
+
+RUN cp modules/c++/nitf/show_nitf++ nitro_Release
+
+RUN polytracker instrument-targets \
```
hbrodin, line 31: We will need to add `--cflog` once #6553 is merged. Additional flag needed to not change existing behaviour.

examples/analysis/ubet/Dockerfile.nitro
```diff
@@ -0,0 +1,61 @@
...
+RUN polytracker build cmake ../.. \
+	-DCMAKE_C_FLAGS="-w -D_POSIX_C_SOURCE=200809L -DCODA_OSS_NO_is_trivially_copyable" \
+	-DCMAKE_CXX_FLAGS="-w -D_POSIX_C_SOURCE=200809L -DCODA_OSS_NO_is_trivially_copyable" \
+  -DCMAKE_BUILD_TYPE=Debug \
+-DCODA_BUILD_TESTS=OFF \
+-DENABLE_PYTHON=OFF
+
+RUN polytracker build cmake  --build . -j$((`nproc`+1)) --clean-first --target show_nitf++ --config Debug
+RUN cp modules/c++/nitf/show_nitf++ nitro_Debug
+RUN polytracker instrument-targets \
```
hbrodin, line 49: We will need to add `--cflog` once #6553 is merged. Additional flag needed to not change existing behaviour.

examples/analysis/ubet/Dockerfile.nitro
```diff
@@ -0,0 +1,61 @@
...
+# and Nitro no longer can find PYTHON_HOME. Using ENABLE_PYTHON=OFF for now
+# so that this Dockerfile at least builds. Enable the Nitro SWIG bindings
+# at your own risk!
+
+RUN apt-get update && \
+	apt-get install -y libcurl4-openssl-dev libssl-dev python3
+
+RUN pip install cxxfilt
+
+RUN git clone https://github.com/mdaus/nitro.git
```
ESultanik, line 16: I suggest we pin to the specific version of Nitro that we know works, just in case something upstream breaks our instrumentation later on.

This commit has I chose is from mid-March, but I didn't test it. @kaoudis do you think we need an earlier version?

```suggestion
RUN git clone https://github.com/mdaus/nitro.git
WORKDIR /polytracker/the_klondike/nitro
RUN git checkout 342f552768e249e86df702062ff3f60ea1ec813a
```

ESultanik (approved):

examples/analysis/ubet/Dockerfile.nitro
kaoudis: I think working on a version that matches the version when we were working on this is probably best for now until I/we get a chance to add more analysis stuff here and start working on this again! 

Looks like some recent commits are just merging changes from coda-oss, which may perhaps be used more inside Maxar (Nitro is the only open source use I am currently aware of) and coda-oss does seem to introduce a good chunk of Nitro's variability. I think pinning to b39ccc4c07e84e6c05cecb9ae24143373a3ed8e2 could be better since that would be the one that we were looking at when we did the work. What do you think?

examples/analysis/ubet/Dockerfile.nitro
kaoudis: Maybe in #6553 we could add `--cflog` here? Or would you prefer to merge #6553 first, then this, with the flag added?

examples/analysis/ubet/Dockerfile.nitro
kaoudis: as above 🙂

examples/analysis/ubet/Dockerfile.nitro
hbrodin: Yes! Let's pin the version you are suggesting @kaoudis! That's the one we have a working result from.

PR #6553, opened 2023-05-17
UBet specific adaptations

hbrodin:
UBet uses a tainted control flow log to get the data needed to detect variability bugs. This is the instrumentation changes that were done to support it.

To enable the `Control Affecting Dataflow` log, invoke the `instrument-targets` command using `--cflog` in addition to the `--taint` and/or `--ftrace` arguments.

There is still some overlap between the features but we can unify that over time. This PR should be non-intrusive in that it shouldn't impact any of the existing functionality.

### DIFF
--- a/polytracker/include/polytracker/passes/tainted_control_flow.h
+++ b/polytracker/include/polytracker/passes/tainted_control_flow.h
@@ -35,12 +35,10 @@ class TaintedControlFlowPass
   void insertTaintStartupCall(llvm::Module &mod);
   void declareLoggingFunctions(llvm::Module &mod);
 
-
   llvm::ConstantInt *get_function_id_const(llvm::Function &f);
   llvm::ConstantInt *get_function_id_const(llvm::Instruction &i);
 
 public:
-
   using function_id = uint32_t;
 
   TaintedControlFlowPass();
--- a/polytracker/include/polytracker/passes/tainted_control_flow.h
+++ b/polytracker/include/polytracker/passes/tainted_control_flow.h
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) 2023-present, Trail of Bits, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed in accordance with the terms specified in
+ * the LICENSE file found in the root directory of this source tree.
+ */
+
+#pragma once
+
+#include <llvm/IR/InstVisitor.h>
+#include <llvm/IR/PassManager.h>
+#include <unordered_map>
+
+namespace polytracker {
+namespace detail {
+struct FunctionMappingJSONWriter;
+}
+
+class TaintedControlFlowPass
+    : public llvm::PassInfoMixin<TaintedControlFlowPass>,
+      public llvm::InstVisitor<TaintedControlFlowPass> {
+  //
+  llvm::IntegerType *label_ty{nullptr};
+  // Taint tracking startup
+  llvm::FunctionCallee taint_start_fn;
+  // Log taint label affecting control flow
+  llvm::FunctionCallee cond_br_log_fn;
+  // Log enter/leave functions
+  llvm::FunctionCallee fn_enter_log_fn;
+  llvm::FunctionCallee fn_leave_log_fn;
+
+  // Helpers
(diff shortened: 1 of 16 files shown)


polytracker/include/taintdag/control_flow_log.h
```diff
@@ -0,0 +1,88 @@
...
+  };
+
+  static constexpr uint8_t tag{8};
+  static constexpr size_t align_of{1};
+  static constexpr size_t allocation_size{1024 * 1024 * 1024};
+
+  template <typename OF>
+  ControlFlowLog(SectionArg<OF> of) : SectionBase(of.range) {}
+
+  void function_event(EventType evt, uint32_t function_id) {
```
surovic, line 49: Is there a reason why we didn't enhance the existing function tracing instead of creating a new section?

surovic (commented):
Disclaimer: I haven't read the PR in full. I got to the instrumentation pass and cflog tdag section. It seems to me like there's a significant overlap between this instrumentation and the existing function tracing, namely with the function entry and exit logging. Am I missing something?

polytracker/include/polytracker/passes/tainted_control_flow.h
```diff
@@ -0,0 +1,66 @@
...
+  void insertTaintStartupCall(llvm::Module &mod);
+  void declareLoggingFunctions(llvm::Module &mod);
+
+  llvm::ConstantInt *get_function_id_const(llvm::Function &f);
+  llvm::ConstantInt *get_function_id_const(llvm::Instruction &i);
+
+public:
+  using function_id = uint32_t;
+
+  TaintedControlFlowPass();
```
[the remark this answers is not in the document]
hbrodin: Me too, but I think in this case we need to have a ctor as it does some initialization (see cpp-file). Otherwise we would have to refactor the code to initialize on first use or something like that. I don't think thats a terribly worthwhile investment, but I could be wrong.

polytracker/include/taintdag/control_flow_log.h
```diff
@@ -0,0 +1,88 @@
...
+#pragma once
+
+#include "taintdag/outputfile.h"
+#include "taintdag/section.h"
+#include "taintdag/taint.h"
+#include "taintdag/util.h"
+
+namespace taintdag {
+
+namespace details {
```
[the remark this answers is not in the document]
hbrodin: agree!

polytracker/include/taintdag/control_flow_log.h
hbrodin: As this was a proof of concept at the time it was written I figured it was easier to create a new one. We weren't really sure what changes we needed. For now, it will allow us to integrate it without breaking anything else. Over time I think the functionality should be merged/unified.

polytracker/include/polytracker/passes/tainted_control_flow.h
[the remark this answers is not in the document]
surovic: I see! Hmm, could we move the initialization of the JSON writer to the `run()` method? That way we can get rid of the default declarations.

polytracker/include/taintdag/control_flow_log.h
surovic: Fair enough. I'll make an issue to keep note of the future merge.

polytracker/include/taintdag/control_flow_log.h
```diff
@@ -0,0 +1,88 @@
...
+// A uint32_t varint encoded by setting highest bit for all but the final byte.
+// Requires up to 5 bytes of storage as each output byte uses 7 input bits.
+// Total maximum need is floor(32/7) = 5. Returns number of bytes required.
+size_t varint_encode(uint32_t val, uint8_t *buffer) {
```
surovic, line 23: Can you explain this encoding scheme to me?

surovic: OK, I found an explanation [here](https://carlmastrangelo.com/blog/lets-make-a-varint). I'm asking myself if this is worth the extra code here. Maybe we could add generic standard compression functionality and compress whole sections instead?

surovic (approved):

PR #6558, opened 2023-07-14
Add guards to instrumentation logging functions

surovic:
This PR adds guards to instrumentation logging functions to logging before all of polytracker's data structures are properly initialized.

### DIFF
--- a/polytracker/src/polytracker/polytracker.cpp
+++ b/polytracker/src/polytracker/polytracker.cpp
@@ -10,7 +10,7 @@
 
 EARLY_CONSTRUCT_EXTERN_GETTER(taintdag::PolyTracker, polytracker_tdag);
 
-bool polytracker_is_initialized = false;
+static bool polytracker_is_initialized = false;
 
 extern "C" taintdag::Functions::index_t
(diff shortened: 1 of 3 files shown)


polytracker/src/polytracker/polytracker.cpp
```diff
@@ -10,22 +10,37 @@
 
 EARLY_CONSTRUCT_EXTERN_GETTER(taintdag::PolyTracker, polytracker_tdag);
 
+static bool polytracker_is_initialized = false;
```
hbrodin, line 13: What is our approach to thread safety? I think we've attempted to keep things thread safe and due to that, we should probably change this into an [atomic_flag](https://en.cppreference.com/w/cpp/atomic/atomic_flag).

hbrodin (commented):
Overall LGTM. The one concern I'd like your input on before approving is regarding thread safety.

polytracker/src/polytracker/polytracker.cpp
surovic: Replaced with `std::atomic_flag`.

hbrodin (approved):

PR #6560, opened 2023-08-16
Rework CI to use action and reduce # of polytracker image builds.

surovic:


.github/workflows/python.yml
```diff
@@ -8,10 +8,12 @@ on:
 
 jobs:
   lint-python:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
+    timeout-minutes: 32
     strategy:
       matrix:
         python-version: ['3.7', '3.8', '3.9', '3.10']
```
ESultanik, line 15: 3.11 is pretty mainstream now, and 3.7 just reached its end of life.
```suggestion
        python-version: ['3.8', '3.9', '3.10', '3.11']
```

ESultanik (approved):
I just have one suggestion, otherwise it looks good to me.

PR #6561, opened 2023-08-16
Format picohttpparser shell script to appease CI

surovic:


examples/http/picohttpparser/example_picohttpparser.sh
```diff
@@ -20,8 +20,7 @@ HOST_DIR=$(dirname "$HOST_PATH")
 # mount the file if it's not already in /workdir
 SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
 if [[ "$HOST_DIR" == "$SCRIPT_DIR" ]]; then
-	docker run --read-only -ti --rm -e POLYPATH="$1" -e POLYDB="$1.tdag" -e POLYTRACKER_STDOUT_SINK=1\
```
lisaoverall, line 23: Looks like I forgot a space here between `1` and `\` at the end of this line in the original, which is probably what made `shfmt` unhappy. You can leave as it is currently, or add the space if we want to enforce code style with shorter line lengths - both are functionally equivalent.

lisaoverall (approved):
Fine by me, modulo your opinion on my comment 🙂

examples/http/picohttpparser/example_picohttpparser.sh
surovic: Less shift-scrolling is always better. Changed back and added the space.

PR #6567, opened 2023-11-14
add nearly-instrumented Imagemagick Dockerfile

kaoudis:
I am not sure if there is somehow a linker error being obscured - I've debugged about five in this attempt to get Imagemagick instrumented - but the last commented out instrumentation command returns non-zero exit status and I am not sure why or what to check next


kaoudis:
Figured out a way to do this and am doing it as part of kaoudis/eval.

PR #6572, opened 2024-08-21
adds currently known publications and public use cases to the README

kaoudis:


README.md
```diff
@@ -199,10 +199,10 @@ instrumentation parameters without needing to recompile the binary.
 ### Environment Variables
 
 PolyTracker accepts configuration parameters in the form of environment
-variables to avoid recompiling target programs. The current environment
-variables PolyTracker supports is:
+variables to avoid recompiling target programs. The current set of environment
+variables that PolyTracker supports is:
 
-```
+```bash
```
hbrodin, line 205: Nit: I guess this isn't really bash... I'd just leave it without the bash language specifier

hbrodin (approved):
LGTM! Just tiny nit that you might want to address.

README.md
kaoudis: the `trunk` tool is upset if we leave it unspecified - I was kind of stumped on what to put. [Looks like](https://github.com/github-linguist/linguist/blob/main/lib/linguist/languages.yml#L6738) we can select from a few different "shell" choices; I think I'll try `shell-script` since that's *sort* of closer heh.

